PR #11, opened 2016-03-01
Add Inverse Kinematics

haudren:
This pull request adds an implementation of Inverse Kinematics. It follows precisely the implementation that was done in [OpenHRP3](https://github.com/fkanehiro/openhrp3/blob/eaf23dc15cdc631e68d38a7db13ec6c9c9c73141/hrplib/hrpModel/InverseKinematics.h).

I am not very happy with the fact that we have to add constants in the .cpp file, but I think that algorithms should work without any tuning: if this is a problem, I can always add accessors to modify them.

Do not hesitate to tell me if the tests and/or doc seems incomplete.

Cheers,
Hervé

### DIFF
--- a/src/IK.cpp
+++ b/src/IK.cpp
@@ -1,0 +1,92 @@
+// This file is part of RBDyn.
+//
+// RBDyn is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// RBDyn is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with RBDyn.  If not, see <http://www.gnu.org/licenses/>.
+
+// associated header
+#include "IK.h"
+
+// includes
+// RBDyn
+#include "FK.h"
+#include "FV.h"
+#include "MultiBody.h"
+#include "MultiBodyConfig.h"
+
+//SpaceVecAlg
+#include <SpaceVecAlg/SpaceVecAlg>
+
+namespace rbd
+{
+
+static const int MAX_IK_ITERATIONS = 50;
+static const double LAMBDA = 0.9;
+static const double IK_THRESHOLD = 1e-8;
+static const double ALMOST_ZERO = 1e-8;
+
+InverseKinematics::InverseKinematics(const MultiBody& mb, int ef_index):
+	ef_index_(ef_index),
+	jac_(mb, mb.body(ef_index).id()),
+	svd_()
+{
+}
+
+struct CwiseRoundOp {
+	CwiseRoundOp(const double& inf, const double& sup) : m_inf(inf), m_sup(sup) {}
+	double operator()(const double& x) const { return x>m_inf && x<m_sup ? 0 : x; }
+	double m_inf, m_sup;
+};
+
+bool InverseKinematics::inverseKinematics(const MultiBody& mb, MultiBodyConfig& mbc,
+                                          const sva::PTransformd& ef_target)
+{
+	int iter = 0;
+	bool converged = false;
+	int dof = 0;
+	rbd::forwardKinematics(mb, mbc);
+	Eigen::MatrixXd jacMat;
+	Eigen::Vector6d v = Eigen::Vector6d::Ones();
+	Eigen::Vector3d rotErr;
+	Eigen::VectorXd res = Eigen::VectorXd::Zero(3);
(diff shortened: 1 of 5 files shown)


src/IK.cpp
```diff
@@ -0,0 +1,103 @@
...
+//SpaceVecAlg
+#include <SpaceVecAlg/SpaceVecAlg>
+
+namespace rbd
+{
+
+static const int MAX_IK_ITERATIONS = 50;
+static const double LAMBDA = 0.9;
+static const double IK_THRESHOLD = 1e-8;
+static const double ALMOST_ZERO = 1e-8;
```
gergondet: This should be in an anonymous namespace to avoid future name conflicts or declare them `static constexpr` in `IK.h` and use as default arguments for the algorithm construction.

src/IK.cpp
```diff
@@ -0,0 +1,103 @@
...
+	svd_()
+{
+}
+
+struct CwiseRoundOp {
+	CwiseRoundOp(const double& inf, const double& sup) : m_inf(inf), m_sup(sup) {}
+	double operator()(const double& x) const { return x>m_inf && x<m_sup ? 0 : x; }
+	double m_inf, m_sup;
+};
+
```
gergondet: Should be in an anonymous namespace

src/IK.cpp
```diff
@@ -0,0 +1,103 @@
...
+		                            ef_target.rotation());
+		v << rotErr, ef_target.translation() - mbc.bodyPosW[ef_index_].translation();
+		converged = v.norm() < IK_THRESHOLD;
+		res = svd_.solve(v);
+
+		dof = 0;
+		for(auto index : jac_.jointsPath())
+		{
+			std::vector<double>& qi = mbc.q[index];
+			for(auto qv=qi.begin(); qv!=qi.end(); ++qv)
```
gergondet: `for(auto & qv : qi)` ?

src/IK.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	int iter = 0;
+	bool converged = false;
+	int dof = 0;
+	rbd::forwardKinematics(mb, mbc);
+	Eigen::MatrixXd jacMat;
+	Eigen::Vector6d v = Eigen::Vector6d::Ones();
+	Eigen::Vector3d rotErr;
+	Eigen::VectorXd res = Eigen::VectorXd::Zero(3);
+	while(iter < MAX_IK_ITERATIONS && ! converged)
```
gergondet: Very minor but
`while( ! converged && iter < MAX_IK_ITERATIONS)`
would read better imho (and perform one less check after convergence but that's very very minor)

src/IK.h
```diff
@@ -0,0 +1,69 @@
...
+class MultiBodyConfig;
+
+/**
+	* Inverse Kinematics algorithm.
+	*/
+class InverseKinematics
+{
+public:
+	InverseKinematics()
+	{}
```
gergondet: Why provide a default constructor?

src/IK.h
```diff
@@ -0,0 +1,69 @@
...
+	/// @param mb MultiBody associated with this algorithm.
+	InverseKinematics(const MultiBody& mb, int ef_index);
+	/**
+	* Compute the inverse kinematics.
+	* @param mb MultiBody used has model.
+	* @param mbc Use q generalized position vector
+	* @return bool if computation has converged
+	* Fill q with new generalized position
```
gergondet: This is slightly inaccurate since you run multiple FK/FV algorithms on `mbc`, `q` is not the only member of `mbc` that is affected by the algorithm. Also if the algorithm fails, `q` and other fields are still filled with whatever is the result of the latest iteration.

gergondet:
Good job!

Since the algorithm can fail, it would be nice to have a failing test included.

CNRS copyright should also be added in IK.h/IK.cpp

src/IK.h
haudren: I must say I copied the default constructor from all other algorithms. It maybe has something to do with how Pybindgen handles copy of C++ objects ?

src/IK.h
haudren: I will modify documentation on this. For the second point, do you think this is a desirable behaviour ? Or should we copy back and forth to a working copy in order to only modify the result if computation has succeeded ?

src/IK.h
gergondet: From a quick test this doesn't seem necessary as long as you don't had the default constructor in the bindings generation (the copy constructor is fine)

src/IK.h
gergondet: I think that anyway the user expects that the `MultiBodyConfig` passed to the algorithm will be modified so this kind of back-tracking should be left to the user. Furthermore, it may be of some interest to get the partial result. It should however be specified in the documentation.

haudren:
@gergondet : I have updated my branch with all of your remarks, except CNRS copyright. Tell me if it looks better now :)

haudren:
Added the copyright, I will probably rebase to squash a few commits before merging!

Issue #10, opened 2016-03-01
Differentiating between indexes and IDs


haudren:
Hello,

A pain point in using SVA/RBDyn is that it is not very clear when and/or why one should use joint (or body) IDs over indexes. For example, even recently @stanislas-brossette found a bug related to computing a jacobian using an index rather than an id.

To alleviate the problem, I started to make id and index two separate classes using boost's `strong_typedef` macro on my [fork](https://github.com/haudren/RBDyn/tree/topic/diff-id-index). However, I am now wondering if IDs should be exposed in the API:
- The main reason to have IDs is to be able to convert configurations between different MultiBodies that have been generated from the same MultiBodyGraph.
- The Jacobian takes an ID as parameter but throws it away immediately
- Most Tasks take an ID to compute a Jacobian but throw it away similarly

Thus, I wonder if would not be simpler to simply drop the IDs from the public API and only expose the indexes.

@jorisv : Is there a fundamental design problem that I missed ?
@bchretien : I know you are using quite a bit of robots created with different roots, do you see any problem with the above ?

Cheers,
Hervé

bchretien:
Well I think that was the first time I mixed up indices/ids, since `mc_rbdyn`'s `createRobotWithBase` expects the base id for the call to `MultiBodyGraph ::makeMultiBody` (see [here](https://github.com/jorisv/RBDyn/blob/1121591371b00e0dd737528f627c29dc5a97091a/src/MultiBodyGraph.h#L170)). Apparently you missed that class on your fork, and that one only deals with ids as far as I can tell.

haudren:
You're right, (my fork is still very much WIP): people who want to build a robot "from scratch" need to deal with ids and indexes. There is simply no other way around it as indexes are created when creating a robot whereas ids are created per-joint. Thus, IDs can be reused across robots with different bases whereas index 0 is _always_ the root joint.

But, once you created the robot, is there really any need to expose the IDs ? I think most users in our case will mainly create a robot and use the resulting `MultiBody` and `MultiBodyConfig` to implement their controller/algorithm.

bchretien:
In order to change the root body, I do need to have access to the ids, so confusion between indices and ids are still possible, even if we were to force the use of indices in other higher-level parts of the API. Thus I'm also in favor of typed ids/indices.

As for consistency in parts of the API that do not have anything to do with `MultiBodyGraph`, it should clearly be addressed, unless there were some good reasons for using both indices and ids (performance?).

[...]

PR #12, opened 2016-03-04
Switch from numerical to textual ID

haudren:
This PR is a follow up to #10 : it is not yet complete, I put it out there for people to be able to review the changes and suggest improvements. I will also probably rebase and squash before merging. It could also be interesting to tag and release the current version, because this will break _all_ libraries relying directly on RBDyn.

In summary, I removed the `int id` property, and equivalently replaced it by `std::string name`. The main thing to watch out when adapting code tot this version is that

``` cpp
mbg.makeMultibody(0, true);
```

Is a valid construct, as `0` can be interpreted as a `const char*` and in turn as a `std::string` but this breaks horribly when run:

> terminate called after throwing an instance of 'std::logic_error'
>  what():  basic_string::_S_construct null not valid
> [1]    15016 abort      ./test_string

Other methods are equally affected, but this one is triggered most often as the base id was almost always zero.

One way to find out those problems is to turn on the zero-as-null-pointer-constant warning from GCC but it also triggers a number of problems within Eigen (In relation with Eigen::Ref apparently). If anyone has good suggestions to avoid this problem ? I sent an e-mail on the Eigen mailing list to get some advice. @gergondet also suggested adding some private versions of the problematic methods that take an `int` in order to break at compilation, but this is quite inelegant.

List of changes:
- [x] Main codebase
- [x] Tests
- [x] Benchmarks
- [x] Python Bindings


gergondet:
This looks good to me (as soon as Benchmarks and bindings are adapted to this change). It may have been nice (to make the transition smoother) to keep the `id()` function and change its signature but this is just pushing back the inevitable.

> @gergondet also suggested adding some private versions of the problematic methods that take an int in order to break at compilation, but this is quite inelegant.

I'm not saying this should stay in the library indefinitely but at least between this release and the next in order to catch such cases easily. This has the advantage of making the API break everywhere it is invoked using int for id instead of string.

Anyway, I'm not sure it's worth worrying about too much as there isn't that many occurrences outside RBDyn:
- Python bindings are probably not affected (since 0 would either disqualify the call or be converted to "0", I'm not sure which pybindgen does)
- Tasks does not build its own MultiBody except in the tests and this will be caught easily (since Tasks is going to be broken anyway and fixed by people who are aware of this issue)
- Same goes for the PG libraries but it's the same case as Tasks 
- Every high-level libraries (i.e. mc_rbdyn_urdf, mc_rtc and mc_ros) pass an int to makeMultiBody and not raw values

haudren:
Thanks for the comments! I adapted the benchmarks (I did not notice any slowdown) and the python bindings (Should pass the travis build). I'll wait to see if there are any other suggestions/questions, and then we'll move on with your plan.

PR #23, opened 2016-11-21
Store mimic joint information

gergondet:
This PR makes it possible to store mimic joint information in a `Joint` object. 

By itself, the information is not very useful in RBDyn and there is currently no way to enforce the mimic relationship in `MultiBodyConfig` for example.

See jrl-umi3218/Tasks#13 for usage in control.

### DIFF
--- a/src/Joint.h
+++ b/src/Joint.h
@@ -107,7 +107,7 @@ public:
 	 * @param multiplier Mimic multiplier
 	 * @param offset Mimic offset
 	 */
-	void set_mimic(const std::string & name, double multiplier, double offset);
+	void setMimic(const std::string & name, double multiplier, double offset);
 
 	/// @return Joint type.
 	Type type() const
@@ -159,27 +159,27 @@ public:
 	}
 
 	/// @return True if the joint is a mimic joint
-	bool is_mimic() const
+	bool isMimic() const
 	{
-		return is_mimic_;
+		return isMimic_;
 	}
 
 	/// @return Mimiced name
-	const std::string& mimic_name() const
+	const std::string& mimicName() const
 	{
-		return mimic_name_;
+		return mimicName_;
 	}
 
 	/// @return Mimic multiplier
-	double mimic_multiplier() const
+	double mimicMultiplier() const
 	{
-		return mimic_multiplier_;
+		return mimicMultiplier_;
 	}
 
 	/// @return Mimic offset
-	double mimic_offset() const
+	double mimicOffset() const
 	{
-		return mimic_offset_;
+		return mimicOffset_;
 	}
 
 	/// @return Joint motion subspace in successor frame coordinate.
@@ -278,10 +278,10 @@ private:
 
 	std::string name_;
 
-	bool is_mimic_ = false;
-	std::string mimic_name_ = "";
-	double mimic_multiplier_ = 1.0;
-	double mimic_offset_ = 0.0;
+	bool isMimic_ = false;
+	std::string mimicName_ = "";
+	double mimicMultiplier_ = 1.0;
+	double mimicOffset_ = 0.0;
 };
 
 
@@ -348,12 +348,12 @@ inline Joint::Joint(Type type,	bool forward, std::string name):
 	constructJoint(type, Eigen::Vector3d::UnitZ());
 }
 
-inline void Joint::set_mimic(const std::string & name, double multiplier, double offset)
+inline void Joint::setMimic(const std::string & name, double multiplier, double offset)
 {
-	is_mimic_ = true;
-	mimic_name_ = name;
-	mimic_multiplier_ = multiplier;
-	mimic_offset_ = offset;
+	isMimic_ = true;
+	mimicName_ = name;
+	mimicMultiplier_ = multiplier;
+	mimicOffset_ = offset;
 }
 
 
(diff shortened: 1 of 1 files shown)


haudren:
This looks pretty good, but I'm unsure on the actual implementation: the fact that we can call `set_mimic()` suggests that we could somehow `clear_mimic()`. I would be in favor of actually having a method `make_mimic(joint, mimic_params)` that returns a new joint, in order to make it clear that there is no way to modify the mimic state once added to the MultiBody. We could also add a new type `MimicJoint : public Joint` but I'm afraid that testing `dynamic_cast<MimicJoint>` has succeeded is too cumbersome.

I will also add a few comments directly in the PR.

src/Joint.h
```diff
@@ -149,6 +158,30 @@ class Joint
 		return name_;
 	}
 
+	/// @return True if the joint is a mimic joint
+	bool is_mimic() const
```
haudren: All methods in RBDyn are camelCase, not snake_case

src/Joint.h
```diff
@@ -149,6 +158,30 @@ class Joint
...
+	{
+		return is_mimic_;
+	}
+
+	/// @return Mimiced name
+	const std::string& mimic_name() const
```
haudren: Same

src/Joint.h
```diff
@@ -149,6 +158,30 @@ class Joint
...
+	{
+		return mimic_name_;
+	}
+
+	/// @return Mimic multiplier
+	double mimic_multiplier() const
```
haudren: same

src/Joint.h
```diff
@@ -149,6 +158,30 @@ class Joint
...
+	{
+		return mimic_multiplier_;
+	}
+
+	/// @return Mimic offset
+	double mimic_offset() const
```
haudren: And one more

src/Joint.h
```diff
@@ -244,6 +277,11 @@ class Joint
 	int dof_;
 
 	std::string name_;
+
+	bool is_mimic_ = false;
+	std::string mimic_name_ = "";
+	double mimic_multiplier_ = 1.0;
+	double mimic_offset_ = 0.0;
```
haudren: All properties in RBDyn are camelCase ending with an underscore, maybe you had too much Python :snake:

src/Joint.h
```diff
@@ -416,7 +462,15 @@ inline sva::MotionVecd Joint::tanAccel(const std::vector<double>& alphaD) const
 
 inline std::vector<double> Joint::zeroParam() const
 {
-	return ZeroParam(type_);
+	auto q = ZeroParam(type_);
```
haudren, line 467: Is using `auto` here really necessary ? Well I guess it's entirely determined by the return type of the function anyways.

src/Joint.h
```diff
@@ -416,7 +462,15 @@ inline sva::MotionVecd Joint::tanAccel(const std::vector<double>& alphaD) const
...
+	if(is_mimic_)
+	{
+		for(auto & qi : q)
+		{
+			qi += mimic_offset_;
```
haudren: What happens if we want to have a non-uniform mimic offset on a non-revolute joint ? mimic_offset should in that case probably be a `std::vector<double>`.

haudren (changes requested):
Some style changes are necessary :(

gergondet:
I've made the changes concerning the case.

Concerning `setMimic` or `makeMimic`, why not meet in the middle by keeping the current implementation but renaming it `makeMimic` to make it clear that the operation cannot be reversed.

I don't think the third option (a new type) would be good. Structures in RBDyn store joints by value and would have to be changed plus the dynamic casting to find out whether a joint is a mimic or not seems like a waste to avoid storing 4 additional values in the structure.

> What happens if we want to have a non-uniform mimic offset on a non-revolute joint ? mimic_offset should in that case probably be a std::vector<double>.

I made myself the same remark when writing the function, I guess this can be added. Note: [in an urdf](http://wiki.ros.org/urdf/XML/joint), this is not an option.

haudren:
As we mostly target URDF, I'd say we can drop it. If someone ever needs it, we can go back to this implementation.

And yes, your suggestion seems sensible, I was only providing the new type as an alternative but it is not desirable.

gergondet:
I've made the change for setMimic and I'm ok with dropping the idea of a vector offset for now.

src/Joint.h
```diff
@@ -100,6 +100,15 @@ class Joint
 		*/
 	Joint(Type type, bool forward, std::string name);
 
+	/**
+	 * Specify mimic information.
```
haudren: Could we just specify in the doc here that this is an irreversible change ?

haudren (changes requested):
Just enhance the doc

haudren (approved):
Looks good to me

src/Joint.h
gergondet: I've changed it to:
```
>-/**
>- * Make the joint mimic and specify mimic information.
>- *
>- * This change cannot be reversed.
>- *
>- * @param name Name of the joint that will be mimiced.
>- * @param multiplier Mimic multiplier
>- * @param offset Mimic offset
>- */
```

PR #25, opened 2017-01-10
MSVC compilation and debug version

aescande:


gergondet:
It looks like you based these changes against the wrong repository. In fact there is no change other than in the cmake folder. Can you work from the current master and update the cmake subtree?

ahundt:
this is super easy to fix just follow github/hub#1339 in a branch checked out from the jrl-umi rbdyn master branch and create a new pull request. followed that for updating #26 to #27

PR #26, opened 2017-01-11
generates standard CMake *Config.cmake and *Version.cmake files

ahundt:
sync with https://github.com/jrl-umi3218/jrl-cmakemodules/pull/103


haudren:
It seems that this PR (similarly to #25 ) was based on the pre-migration version of RBDyn. Long story short, we migrated the main repository from `jorisv` to `jrl-umi3218`, but the commit history was modified and conflicts. Can you base your PR on this version ? I think you basically only want to update cmake and slightly modify `CMakeLists.txt`.

ahundt:
@haudren oh geez that's scary... hmm it will be tough for me to go back on all my repos that are based on the jorisv versions... 

possible ways around this:
https://github.com/blog/2141-squash-your-commits
https://help.github.com/articles/about-pull-request-merges/
https://github.com/github/hub/issues/1339

ahundt:
That last one github/hub#1339 worked beautifully and now everything is in #27 so closing this.

PR #31, opened 2017-04-26
Fix python bindings

haudren:
This is a just a typo fix: I forgot to remove a reference to "Id" in the legacy python bindings during the great ID removal.

I open this pull request in case something else comes up and needs to be added.


ahundt:
@haudren @gergondet this looks like it should still be merged, doesn't it?

gergondet:
That's the old bindings actually. Closing this PR. The function has the correct name in the current bindings.

PR #33, opened 2017-05-25
Using acceleration in the integration of joint position.

aescande:
This PR basically corrects the EulerIntegration scheme.
Previously, it was done like this:
```
x(k+1) = x(k) + v(k)*dt
v(k+1) = v(k) + a(k)*dt
```
or corresponding formulas for more complex joints. (`x` is the joint position, `v` its velocity and `a` its acceleration). This was corresponding to integrate position with constant velocity then velocity with constant acceleration, so that the acceleration `a(k)` at instant `k` was only influencing the position `x(k+2)`.
Now it is done like this:
```
x(k+1) = x(k) + v(k)*dt + 1/2*a(k)*dt^2
v(k+1) = v(k) + a(k)*dt
```
which corresponds to integrating both position and velocity for a constant acceleration.

There are still things to check/correct:
- for planar joints, I kept the old formula, because I don't quite understand (yet) what is done.
- the integration of the translation part of a free joint seems odd: integrating N time a constant velocity with increments dt/N or once with increment dt does not yield the same results. This was already the case in previous version.

### DIFF
--- a/src/EulerIntegration.cpp
+++ b/src/EulerIntegration.cpp
@@ -25,7 +25,31 @@
 
 namespace
 {
-	/** Compute the sum of the first terms of the Magnus expansion of \Omega such that
+	namespace detail
+	{
+		double constexpr sqrtNewtonRaphson(double x, double curr, double prev)
+		{
+			return curr == prev
+				? curr
+				: sqrtNewtonRaphson(x, 0.5 * (curr + x / curr), curr);
+		}
+
+		/**
+		* Constexpr version of the square root
+		* Return value:
+		*   - For a finite and non-negative value of "x", returns an approximation for the square root of "x"
+		*   - Otherwise, returns NaN
+		* Copied from https://stackoverflow.com/a/34134071
+		*/
+		double constexpr sqrt(double x)
+		{
+			return x >= 0 && x < std::numeric_limits<double>::infinity()
+				? sqrtNewtonRaphson(x, x, 0)
+				: std::numeric_limits<double>::quiet_NaN();
+		}
+	}
+
+		/** Compute the sum of the first terms of the Magnus expansion of \Omega such that
 		* q' = q*exp(\Omega) is the quaternion obtained after applying a constant acceleration
 		* rotation wD for a duration step, while starting with a velocity w
 		*/
@@ -41,14 +65,14 @@ namespace
 		return O1 + O2 + O3;
 	}
 
-	/** sinus cardinal: sin(x)/x 
+		/** sinus cardinal: sin(x)/x
 		* Code adapted from boost::math::detail::sinc
 		*/
 	double sinc(double x)
 	{
-		const double taylor_0_bound = std::numeric_limits<double>::epsilon();
-		const double taylor_2_bound = sqrt(taylor_0_bound);
-		const double taylor_n_bound = sqrt(taylor_2_bound);
+		constexpr double taylor_0_bound = std::numeric_limits<double>::epsilon();
+		constexpr double taylor_2_bound = detail::sqrt(taylor_0_bound);
+		constexpr double taylor_n_bound = detail::sqrt(taylor_2_bound);
 
 		if (std::abs(x) >= taylor_n_bound)
 		{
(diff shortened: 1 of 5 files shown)


haudren:
Quick remark for the test:
```
warning: enumeration value ‘Fixed’ not handled in switch [-Wswitch]
```

haudren:
So something is a bit fishy, the tests pass on Windows, but fail on Linux... I wonder where the difference lies.

aescande:
Yes, and it does not seems to be small numeric differences. I'll investigate that

haudren:
Might have something to do with this:
```
warning: conversion to ‘int’ from ‘double’ may alter its value [-Wconversion]
     if (abs(x) >= taylor_n_bound)
```

haudren:
Ok so if you replace `abs` with `fabs` the tests still fail but with tiny errors:
```
Running 9 test cases...
/home/herve/dev/RBDyn/tests/AlgoTest.cpp(467): error in "EulerTest": check { q.begin(), q.end() } == { goalQ.begin(), goalQ.end() } failed. 
Mismatch in a position 1: 0.707107 != 0.707107

*** 1 failure detected in test suite "Algo"
```
And
```
Running 1 test case...
/home/herve/dev/RBDyn/tests/IntegrationTest.cpp(189): error in "JointIntegrationTest": difference{0.0010542} between mbc0.q[1][i]{0.44383554716773121} and mbc.q[1][i]{0.44336814831097465} exceeds 0.001

*** 1 failure detected in test suite "JointTest"
```

gergondet:
I'm confirming Hervé remark on the test. Switching to `std::abs` (and `std::sqrt` although it has no effect) mostly fixes the tests. The error can be fixed by reducing the time step (actually, the test passes even with 1100 instead of 1000).

Edit: just saw you pushed a fix, so... cool :+1:

gergondet:
The PR is fine imho but you totally broke the indentation as RBDyn uses tabs and not spaces (you also left a few trailing spaces which is not as big a deal).

Can you fix it easily with Visual Studio? (otherwise if you give me write access to your fork I'll fix the indentation for you)

Also, poke @jorisv concerning planar joints

haudren:
BTW this breaks the Tasks Mimic joint test. You will need to increase the stiffness of the task to make them pass.

src/EulerIntegration.cpp
```diff
@@ -23,84 +23,140 @@
...
-Eigen::Matrix<double, 4, 3> angularVelToQuatVel(const std::vector<double>& q)
+namespace
 {
-	return 0.5*(Eigen::Matrix<double, 4, 3>() <<
-		-q[1], -q[2], -q[3],
-		q[0], -q[3], q[2],
-		q[3], q[0], -q[1],
-		-q[2], q[1], q[0]).finished();
-}
+	/** Compute the sum of the first terms of the Magnus expansion of \Omega such that
```
gergondet: Looks strange depending on your tab space setting. Please add a tab in front of the comment start

src/EulerIntegration.cpp
```diff
@@ -23,84 +23,140 @@
...
+		Eigen::Vector3d O3 = wD.cross(O2) * step2 / 20;
 
+		return O1 + O2 + O3;
+	}
 
-void eulerJointIntegration(Joint::Type type, const std::vector<double>& alpha,
-	double step, std::vector<double>& q)
-{
-	switch(type)
+	/** sinus cardinal: sin(x)/x
```
gergondet: Same

tests/IntegrationTest.cpp
```diff
@@ -0,0 +1,220 @@
...
+		v = randVec(2, -1, 1);
+		a = randVec(2, -1, 1);
+		break;
+	case Joint::Free:
+		q = randVec(4, -1, 1, true);
+		auto qt = randVec(3, -1, 1);
+		q.insert(q.end(), qt.begin(), qt.end());
+		v = randVec(6, -1, 1);
+		a = randVec(6, -1, 1);
+		break;
```
gergondet, line 123: For the sake of silencing the warning here:
```cpp
default:
  break;
```

gergondet (changes requested):
Only cosmetic changes and we're good to go :+1:

gergondet (approved):

PR #39, opened 2017-06-26
add functions to map param or dof indices to joint indices

barcode:
i wrote the function since i needed it in some cases, but i don't need it any more.  
but maybe someone else does need it.

### DIFF
--- a/src/RBDyn/MultiBody.h
+++ b/src/RBDyn/MultiBody.h
@@ -22,6 +22,7 @@
 #include <stdexcept>
 #include <unordered_map>
 #include <vector>
+#include <algorithm>
 
 // SpaceVecAlg
 #include <SpaceVecAlg/SpaceVecAlg>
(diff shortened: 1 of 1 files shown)


haudren:
I'm not sure that this is really useful in general, what was your use-case? I don't see anything wrong but I'd rather not add too many functions to RBDyn as it might get confusing for end-users.

As a side-note, `std::upper_bound` is not `O(1)`, contrarily to other functions like `jointPosInParam` and the likes. Should those functions be added, we would need to mention it in the documentation.

barcode:
it was some debugging code where i had all pos values/velocities in one big vector (which was a stupid idea) and hat do do this mapping.

haudren:
Ok, but wouldn't this be alleviated by using `rbd::paramToVector` and the likes? They translate from the `Eigen::VectorXd` into a `std::vector<std::vector<double>>`, and you then index by joint.

barcode:
it would have if i had known these functions (they are not in the mb interface so i did not see them).  
since i only needed it temporarily i went for the quick version and wrote this.  
i do not need these functions, but thought maybe they could be of use to someone.

PR #42, opened 2017-11-06
README.md cython bindings dependency update

ahundt:


gergondet:
Same deal as jrl-umi3218/SpaceVecAlg#20 regarding the CI break.

This PR will then also fix the CI. Will merge when it succeeds.

Thanks for the patch :cake:

PR #44, opened 2017-12-08
setup.in.py use sys.platform instead of os.name and add darwin check

ahundt:
This is a simpler and more reliable way to check the OS version in python. Also fixes MacOS install bug detailed in https://github.com/jrl-umi3218/SpaceVecAlg/issues/16 which also applies here in RBDyn.


gergondet:
Hi Andrew, same as jrl-umi3218/SpaceVecAlg#21

Sorry for the trouble but I'll land a patch that removes the python-config call entirely

PR #46, opened 2018-01-09
Topic/add full jacobian

wuv-ogre:
added addFullJacobian() and sAddFullJacobian()
changed Momentum.cpp to use addFullJacobian()
which overal removes the adding of many zeros

Issue #45, opened 2018-01-03
efficiency of CentroidalMomentumMatrix::computeMatrixAndMatrixDot etc.


wuv-ogre:
If I'm not mis-understanding, code like

jacWork_[i] = proj*jac;
jacVec_[i].fullJacobian(mb, jacWork_[i], jacFull_);
cmMat_.block(0, 0, 6, mb.nrDof()) += jacFull_;

could turn into

            auto add_in = []( const MultiBody& mb, const std::vector<int>& jointsPath, Eigen::MatrixXd& cmMat, Eigen::MatrixXd& jac )
            {
                int jacPos = 0;
                for(std::size_t index = 0; index < jointsPath.size(); ++index)
                {
                    int j = jointsPath[index];
                    int dof = mb.joint(j).dof();
                    
                    cmMat.block(0, mb.jointPosInDof(j), jac.rows(), dof) += jac.block(0, jacPos, jac.rows(), dof);
                    
                    jacPos += dof;
                }
            };
            
            const std::vector<int>& jointsPath = jacVec_[i].jointsPath();
            jacWork_[i] = proj*jac;
            add_in( mb, jointsPath, cmMat_, jacWork_[i] );

doing so bumped my frame rate from about 7 to 20 fps for a 50 something dof multibody

gergondet:
Hi @wuv-ogre 

This is indeed a nice improvement. That's pretty much the implementation of `fullJacobian` but the result is accumulated rather than zero-ed. I think it is relevant to add this function in the `Jacobian` class (something like `addFullJacobian`) and use it in RBDyn (although it also seems to be only useful in `Momentum.cpp`)

Would you have time to submit a PR?

wuv-ogre:
done a quick check with 

void Jacobian::addFullJacobian(const MultiBody& mb,
                         const Eigen::Ref<const Eigen::MatrixXd>& jac,
                         Eigen::MatrixXd& res) const
{
    int jacPos = 0;
    for(std::size_t index = 0; index < jointsPath_.size(); ++index)
    {
        int i = jointsPath_[index];
        int dof = mb.joint(i).dof();
        res.block(0, mb.jointPosInDof(i), jac.rows(), dof) +=
            jac.block(0, jacPos, jac.rows(), dof);
        jacPos += dof;
    }
}

and the momentum code going to

            jacWork_[i] = proj*jac;
[...]

PR #49, opened 2018-05-01
Add coriolis matrix computation

haudren:
This PR adds the work I did while I was still at the lab: it allows to compute the full Coriolis matrix (i.e. `C(q, qd)` instead of `C(q, qd)qd`).

I know it is not perfect, but I would like to put it out there !

There might also be some overlap with #46 as I had to implement a "block expand" functionality to obtain decent performance.

Things to fix:
- [x] Indentation should use tabs
- [x] Tests should be a real test, not just a print
- [x] Windows build is probably broken

Maybes (what's your opinion @gergondet ):
- [x] Make a Coriolis benchmark
- [x] Split the Jacobian expand logic from the Coriolis side
- [x] Add python bindings


haudren:
I used the following `clang-format` config to fix my indentation, might be useful to put it somewhere in the project.

```yaml
BasedOnStyle: LLVM
IndentWidth: 8
UseTab: Always
BreakBeforeBraces: Allman
AllowShortIfStatementsOnASingleLine: false
IndentCaseLabels: false
```

haudren:
@gergondet : Any remarks ?

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
+#ifndef CORIOLIS_H_KED6CAHK
+#define CORIOLIS_H_KED6CAHK
```
gergondet: Please use `#pragma once`

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
+#ifndef CORIOLIS_H_KED6CAHK
```
gergondet: Missing copyright header

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
...
+#define CORIOLIS_H_KED6CAHK
+
+#include <array>
+
+#include <RBDyn/MultiBodyConfig.h>
+#include <RBDyn/CoM.h>
+#include <rbdyn/config.hh>
```
gergondet: Not needed since it is included by others

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace rbd
+{
+
+using Block = std::array<int, 3>;
```
gergondet: `Eigen::DenseIndex` or `Eigen::Index` (>= 3.3)

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
...
+using Blocks = std::vector<Block>;
+
+/** Expand a symmetric product of a jacobian by its transpose onto every DoF.
```
gergondet: I'm not sure I understand what you mean about the symmetric product here, `jacMat` can be anything.

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
...
+RBDYN_DLLAPI Blocks compactPath(const rbd::Jacobian& jac,
+		const rbd::MultiBody& mb);
+
+/** Expand a symmetric product of a jacobian by its transpose onto every DoF
+ * and accumulate the result using a compact representation of the DoF.
+ * @param compacPath Blocks representing the compact kinematic path of the Jacobian
+ * @param jacMat The matrix containing the product J^T*J to be expanded
+ * @param res The accumulator matrix
+ **/
+RBDYN_DLLAPI void compactExpandAdd(const Blocks& compactPath,
```
gergondet: Rename `expandAdd` and let C++ overloading do its magic

src/RBDyn/Coriolis.h
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Computation of the Coriolis effects matrix on a multibody. The Coriolis
+ * factorization is not unique, we use the formulation of M. Bjerkend and
+ * K. Pettersen in "A new Coriolis matrix factorization", 2012
+ * NB: ForwardDynamics::C() directly computes the product of this matrix with qd.
+ * This C*qd is unique, but C itself is not.
+ */
+class RBDYN_DLLAPI Coriolis
+{
+	public:
```
gergondet: `public`/`private` alignment should be the same as the class declaration

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <RBDyn/EulerIntegration.h>
+#include <RBDyn/FD.h>
+#include <RBDyn/FK.h>
+#include <RBDyn/FV.h>
+
+#include <iostream>
+
+BOOST_AUTO_TEST_CASE(CoriolisTest)
+{
+	std::srand(133757348);
```
gergondet: `std::srand(std::time(0));` looks [ok on VS](https://godbolt.org/g/P6gqeT)

You should also `#include <cstdlib>` and `#include <ctime>`

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+	rbd::MultiBodyConfig mbc;
+	rbd::MultiBodyGraph mbg;
+
+	std::tie(mb, mbc, mbg) = makeTree30Dof(false);
+
+	rbd::MultiBody mbt;
+	rbd::MultiBodyConfig mbtc;
+	rbd::MultiBodyGraph mbtg;
+
+	std::tie(mbt, mbtc, mbtg) = makeTree30Dof(false);
```
gergondet: Why do you bother with having the same robot twice? This does not look necessary and it makes the test code a bit tedious to read

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+				Eigen::Vector3d axis =
+				    Eigen::Vector3d::Random();
+				Eigen::AngleAxisd aa(0.5, axis / axis.norm());
+				Eigen::Quaterniond qd(aa);
+				q[0] = qd.w();
+				q[1] = qd.x();
+				q[2] = qd.y();
+				q[3] = qd.z();
+			}
+		}
```
gergondet: This could be a small function. Also, you can assume the free-flyer is `mbc.q[0]` and avoid the loop/check here

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		mbtc.q = mbc.q;
+
+		rbd::forwardKinematics(mb, mbc);
+		rbd::forwardVelocity(mb, mbc);
+
+		rbd::forwardKinematics(mbt, mbtc);
+		rbd::forwardVelocity(mbt, mbtc);
+
+		rbd::ForwardDynamics fd(mbt);
```
gergondet: You can keep this object out of the loop

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+	const static int ROUNDS = 1000;
+
+	for (int i = 0; i < ROUNDS; ++i)
+	{
+		mbc.zero(mb);
+		mbtc.zero(mbt);
+
+		Eigen::VectorXd q = Eigen::VectorXd::Random(mb.nrParams());
+		mbc.q = rbd::sVectorToParam(mb, q);
```
gergondet: No need for the safe version here

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		rbd::forwardKinematics(mbt, mbtc);
+		rbd::forwardVelocity(mbt, mbtc);
+
+		rbd::ForwardDynamics fd(mbt);
+		fd.computeC(mbt, mbtc);
+		Eigen::VectorXd gravity = fd.C();
+
+		Eigen::VectorXd qd = Eigen::VectorXd::Random(mb.nrDof());
+		mbc.alpha = rbd::sVectorToDof(mb, qd);
```
gergondet: No need for the safe version here

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+		mbtc.alpha = mbc.alpha;
+
+		rbd::forwardVelocity(mb, mbc);
+		rbd::forwardVelocity(mbt, mbtc);
+
+		rbd::Coriolis coriolis(mb);
```
gergondet: You can keep this object out of the loop

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+		Eigen::MatrixXd N = fd.C();
+
+		BOOST_CHECK_SMALL((C * qd + gravity - N).norm(), TOL);
+
+		fd.computeH(mbt, mbtc);
+		Eigen::MatrixXd m1 = fd.H();
+
+		double dt = 1e-8;
+
+		rbd::sEulerIntegration(mbt, mbtc, dt);
```
gergondet: No need for the safe version here

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		fd.computeC(mbt, mbtc);
+		Eigen::MatrixXd N = fd.C();
+
+		BOOST_CHECK_SMALL((C * qd + gravity - N).norm(), TOL);
+
+		fd.computeH(mbt, mbtc);
+		Eigen::MatrixXd m1 = fd.H();
+
+		double dt = 1e-8;
```
gergondet: This is a really small timestep, why?

tests/ExpandTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <RBDyn/FD.h>
+#include <RBDyn/FK.h>
+#include <RBDyn/FV.h>
+#include <RBDyn/EulerIntegration.h>
+
+#include <iostream>
+
+BOOST_AUTO_TEST_CASE(ExpandJacobianTest)
+{
+	std::srand(133757348);
```
gergondet: Same comment as in the Coriolis test

tests/CoriolisTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+#define BOOST_TEST_MODULE Coriolis
+#include <boost/test/unit_test.hpp>
+
+#include <RBDyn/Coriolis.h>
+#include <RBDyn/EulerIntegration.h>
+#include <RBDyn/FD.h>
+#include <RBDyn/FK.h>
+#include <RBDyn/FV.h>
+
+#include <iostream>
```
gergondet: Not needed

tests/ExpandTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#define BOOST_TEST_MODULE Expand
+#include <boost/test/unit_test.hpp>
+
+#include <RBDyn/FD.h>
+#include <RBDyn/FK.h>
+#include <RBDyn/FV.h>
+#include <RBDyn/EulerIntegration.h>
+
+#include <iostream>
```
gergondet: Not needed

tests/ExpandTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+	rbd::Jacobian jac(mb, mb.body(mb.nrBodies() - 1).name());
+	rbd::Blocks compact = rbd::compactPath(jac, mb);
+
+	for(int i = 0; i < ROUNDS; ++i)
+	{
+		mbc.zero(mb);
+
+		Eigen::VectorXd q = Eigen::VectorXd::Random(mb.nrParams());
+		mbc.q = rbd::sVectorToParam(mb, q);
```
gergondet: No need for the safe version here

tests/ExpandTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+			{
+				Eigen::Vector3d axis = Eigen::Vector3d::Random();
+				Eigen::AngleAxisd aa(0.5, axis/axis.norm());
+				Eigen::Quaterniond qd(aa);
+				q[0] = qd.w();
+				q[1] = qd.x();
+				q[2] = qd.y();
+				q[3] = qd.z();
+			}
+		}
```
gergondet: Same remark as Coriolis

tests/ExpandTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <RBDyn/Coriolis.h>
+
+#define BOOST_TEST_MODULE Expand
+#include <boost/test/unit_test.hpp>
+
+#include <RBDyn/FD.h>
+#include <RBDyn/FK.h>
+#include <RBDyn/FV.h>
+#include <RBDyn/EulerIntegration.h>
```
gergondet: You only need/use FK/FV here

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+	Eigen::MatrixXd res = Eigen::MatrixXd::Zero(mb.nrDof(), mb.nrDof());
+	expandAdd(jac, mb, jacMat, res);
+	return res;
+}
+
+void expandAdd(const rbd::Jacobian& jac,
+			const rbd::MultiBody& mb,
+			const Eigen::MatrixXd& jacMat,
+			Eigen::MatrixXd& res)
+{
```
gergondet: Needs an assert on the size of `res`

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+std::vector<std::array<int, 3>> compactPath(const rbd::Jacobian& jac,
+						const rbd::MultiBody& mb)
+{
+	std::vector<std::array<int, 3>> res;
+
+	int start_block = mb.jointPosInDof(jac.jointsPath()[0]);
+	int len_block = mb.joint(jac.jointsPath()[0]).dof();
+
+	int start;
```
gergondet: Declare inside the loop since it is not meant to be used outside

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+	int startJac = 0;
+
+	for(int i : jac.jointsPath())
+	{
+		if(i == jac.jointsPath()[0])
+		{
+			continue;
+		}
```
gergondet: Replace with the more old-school but more explicit in this case:
```cpp
const auto & jPath = jac.jointsPath();
for(size_t j = 1; j < jPath().size(); ++j)
{
    int i = jPath[j];
```

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+		jacs_.push_back(rbd::Jacobian(mb, mb.body(i).name(), com));
+		jacMats_.push_back(Eigen::MatrixXd(6, jacs_[i].dof()));
+		jacDotMats_.push_back(Eigen::MatrixXd(6, jacs_[i].dof()));
+		compactPaths_.push_back(compactPath(jacs_[i], mb));
+	}
+}
+
+Eigen::MatrixXd Coriolis::coriolis(const rbd::MultiBody& mb, const rbd::MultiBodyConfig& mbc)
+{
+	Eigen::MatrixXd coriolis = Eigen::MatrixXd::Zero(mb.nrDof(), mb.nrDof());
```
gergondet: We probably want to keep this allocated in the `Coriolis` object and change this function declaration to return a const reference to it.

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+	Eigen::Matrix3d rot;
+	Eigen::Matrix3d rDot;
+
+	Eigen::Matrix3d inertia;
+
+	for(int i = 0; i < mb.nrBodies(); ++i)
+	{
+		jacMats_[i] = jacs_[i].jacobian(mb, mbc);
+
+		jacDotMats_[i] = jacs_[i].jacobianDot(mb, mbc);
```
gergondet: I don't think you need to have `jacMats_` and `jacDotMats_`, `Jacobian::jacobian` and `Jacobian::jacobianDot` do the work on the matrix they store and returns you a reference to that matrix so you could have:

```cpp
const auto & jac = jacs_[i].jacobian(mb, mbc);
const auto & jacDot = jacs_[i].jacobianDot(mb, mbc);
```

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+		auto jvi = jacMats_[i].bottomRows<3>();
+		auto jDvi = jacDotMats_[i].bottomRows<3>();
+
+		auto jwi = jacMats_[i].topRows<3>();
+		auto jDwi = jacDotMats_[i].topRows<3>();
+
+		double mass = mb.body(i).inertia().mass();
+		inertia = mb.body(i).inertia().inertia()
+			- sva::vector3ToCrossMatrix<double>(mass*jacs_[i].point())*sva::vector3ToCrossMatrix(jacs_[i].point()).transpose();
```
gergondet: For efficiency sake you might want a specific function to return `sva::vector3ToCrossMatrix<double>(point)*sva::vector3ToCrossMatrix<double>(point).transpose()` directly

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+		Eigen::Matrix3d ir = inertia*rot.transpose();
+
+		/* C = \sum m_i J_{v_i}^T \dot{J}_{v_i}
+		 *        + J_{w_i}^T R_i I_i R_i^T \dot{J}_{w_i}
+		 *        + J_{w_i}^T \dot{R}_i I_i R_i^T J_{w_i} */
+
+		Eigen::MatrixXd res = mass*jvi.transpose()*jDvi
+				+ jwi.transpose()*((rot*ir)*jDwi
+				+  (rDot*ir)*jwi);
```
gergondet: Please don't cut the line at this point, it makes it look like this does not belong in the parentheses expression

src/Coriolis.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+		compactExpandAdd(compactPaths_[i], res, coriolis);
```
gergondet: If `compactExpandAdd` was accepting `const Eigen::Ref<const Eigen::MatrixXd> &` instead of `const Eigen::MatrixXd &` we could allocate `res` once (at the biggest jacobian size) and pass a `res` block here. This should be benched though...

gergondet (changes requested):

gergondet:
Sorry for the long delay, I was on vacation but I am back to work now!

> Make a Coriolis benchmark

Sure.

>  Split the Jacobian expand logic from the Coriolis side

Sure. That could be use as in #46 

Few things that I didn't add to the review:
- This could be part of the `Jacobian` interface
- We could store the compact path in the `Jacobian`
- I would use a light structure rather than a raw array for `Block`, this make `compactExpandAdd` hard to decipher without looking at how the compact path is computed

>  Add python bindings

Will do when the C++ version is finalized.

tests/CoriolisTest.cpp
haudren: I just wanted the test to be random, but reproducible, hence the fixed seed. If you prefer a truly random seed, I'll change it back.

tests/CoriolisTest.cpp
haudren: Because we will compute finite differences on H, not perform integration for simulation. Hence, the timestep must be as small as possible.

src/RBDyn/Coriolis.h
haudren: Well `jacMat` can be anything, but it *must* be a `dof x dof` matrix. Then, it will be expanded by blocks of `dof_j1 x dof_j2` which means that a `jacMat` not representing a `J^T J`  (symmetric product) is meaningless.

tests/CoriolisTest.cpp
haudren: Indeed. This is a leftover from when I was trying various settings (with / without inertia transformed to origin, with / without floating base...).

src/RBDyn/Coriolis.h
haudren: Done

src/RBDyn/Coriolis.h
haudren: Done

src/RBDyn/Coriolis.h
haudren: Done

src/RBDyn/Coriolis.h
haudren: Done

src/RBDyn/Coriolis.h
haudren: Done

src/RBDyn/Coriolis.h
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/ExpandTest.cpp
haudren: Done

tests/CoriolisTest.cpp
haudren: Done

tests/ExpandTest.cpp
haudren: Done

tests/ExpandTest.cpp
haudren: Done

tests/ExpandTest.cpp
haudren: Done

tests/ExpandTest.cpp
haudren: Done

src/Coriolis.cpp
haudren: Done

src/Coriolis.cpp
haudren: Done

src/Coriolis.cpp
haudren: Fixed it, thanks

src/Coriolis.cpp
haudren: Done

src/Coriolis.cpp
haudren: Done

src/Coriolis.cpp
haudren: Done

haudren:
@gergondet : I think I fixed a lot of the mentioned points, but I'm out of time for today, I'll come back to it tomorrow to fix the rest. Feel free to modify the branch as you see fit of course.

haudren:
Note that the build is failing because it can't find boost on Windows apparently? Weird.

src/Coriolis.cpp
haudren: Do you really think we would gain much by splitting this into a function?

haudren:
I tried implementing your suggestion of using only one `res` and passing blocks of it:

Version            | Debug (100 rounds) | Release (1000 rounds)
------------------|--------------------------|-----------------------------
One allocation |  7.75s  |  0.13s
Multiple allocs  | 7.94s  | 0.15s

~~So it is *slower* to have a single allocation in Release apparently, the compiler must be doing some magic behind the scenes.~~

It is marginally faster to have a single allocation in Release.

here is the diff:
```diff
diff --git a/src/Coriolis.cpp b/src/Coriolis.cpp
index 8c2243f..1bdbcc8 100644
--- a/src/Coriolis.cpp
+++ b/src/Coriolis.cpp
@@ -62,8 +62,8 @@ Blocks compactPath(const rbd::Jacobian& jac,
 }
 
 void expandAdd(const Blocks& compactPath,
-			const Eigen::MatrixXd& jacMat,
-			Eigen::MatrixXd& res)
+		const Eigen::Ref<const Eigen::MatrixXd>& jacMat,
+		Eigen::MatrixXd& res)
 {
 	for(const auto& b1 : compactPath)
 	{
@@ -105,6 +105,14 @@ const Eigen::MatrixXd& Coriolis::coriolis(const rbd::MultiBody& mb, const rbd::M
 
         coriolis_.setZero();
 
+        int res_size = 0;
+        for(const auto& jac : jacs_)
+        {
+          res_size = std::max(res_size, jac.dof());
+        }
+
+        Eigen::MatrixXd res = Eigen::MatrixXd::Zero(res_size, res_size);
+
 	for(int i = 0; i < mb.nrBodies(); ++i)
 	{
 		const auto & jac = jacs_[i].jacobian(mb, mbc);
@@ -129,10 +137,10 @@ const Eigen::MatrixXd& Coriolis::coriolis(const rbd::MultiBody& mb, const rbd::M
 		 *        + J_{w_i}^T R_i I_i R_i^T \dot{J}_{w_i}
 		 *        + J_{w_i}^T \dot{R}_i I_i R_i^T J_{w_i} */
 
-		Eigen::MatrixXd res = mass*jvi.transpose()*jDvi
+		res.topLeftCorner(jacs_[i].dof(), jacs_[i].dof()) = mass*jvi.transpose()*jDvi
 				+ jwi.transpose()*((rot*ir)*jDwi +  (rDot*ir)*jwi);
 
-		expandAdd(compactPaths_[i], res, coriolis_);
+		expandAdd(compactPaths_[i], res.topLeftCorner(jacs_[i].dof(), jacs_[i].dof()), coriolis_);
 	}
 
 	return coriolis_;
diff --git a/src/RBDyn/Coriolis.h b/src/RBDyn/Coriolis.h
index d9512c4..904cd5e 100644
--- a/src/RBDyn/Coriolis.h
+++ b/src/RBDyn/Coriolis.h
@@ -73,7 +73,7 @@ RBDYN_DLLAPI Blocks compactPath(const rbd::Jacobian& jac,
  * @param res The accumulator matrix
  **/
 RBDYN_DLLAPI void expandAdd(const Blocks& compactPath,
-			const Eigen::MatrixXd& jacMat,
+			const Eigen::Ref<const Eigen::MatrixXd>& jacMat,
 			Eigen::MatrixXd& res);
 
 /**
diff --git a/tests/CoriolisTest.cpp b/tests/CoriolisTest.cpp
index 636d605..a1a1cc5 100644
--- a/tests/CoriolisTest.cpp
+++ b/tests/CoriolisTest.cpp
@@ -38,7 +38,7 @@ BOOST_AUTO_TEST_CASE(CoriolisTest)
 	rbd::ForwardDynamics fd(mb);
 	rbd::Coriolis coriolis(mb);
 
-	const static int ROUNDS = 100;
+	const static int ROUNDS = 1000;
 
 	for (int i = 0; i < ROUNDS; ++i)
 	{
```

src/Coriolis.cpp
haudren: See my comment.

haudren:
My bad, I forgot to change the number of rounds in the Release/Multiple Allocs case. Updated the table.

src/Coriolis.cpp
gergondet: Ok, I've done a similar patch in 6db09a6fc54bba54c2ceba028efb7ff46cffa48 plus a small improvement in expand functions (using `noalias()` on `res` which seems reasonable)

gergondet:
I've adressed the 3 remaining todos and used the `Block` feature in Momentum, this should now close #46 and #45 

Actually, this opens the possibility for a very small optimization in the computation of the momentum dot matrix.

I will try to figure out why the AppVeyor build is failing tomorrow...

haudren:
@gergondet : Nice work, thanks ! Let me know if you need anything else :+1:

haudren:
I also made a quick bench for momentum:

Revision | Momentum matrix | Momentum matrix dot |
------------|------------------------|------------------------------|
a1e6f0ad (master) | 5.6e-5 | 1.2e-4 |
56b5449b (PR) | 2.7e-5 | 5.5e-5| 

So we are on average about twice as fast, not bad :smile: 

@gergondet : Should I commit this benchmark in `JacobianBench` ?

gergondet (approved):

gergondet:
Ok, CI should be ok now after updating the Boost version.

Will merge when all CI lands.

haudren:
Thanks :fireworks:

PR #50, opened 2018-06-28
Topic/id factorization

vsamy:
Factorization of the ID algorithm to allow ID without taking into account the inertial parameters.


gergondet:
API wise I'd rather see two functions:
- `inverseDynamics(...)` the current function
- `inverseDynamicNoInertia(...)` which computes the ID without taking the inertia into account

The bindings code is broken in this PR but with this API it would be irrelevant.

gergondet (approved):

gergondet:
Great! Thanks :cake:

Issue #65, opened 2019-07-03
Discrepencies in conventions between planar and free joints


aescande:
In Featherstone's book, there is two competing conventions for representing the translation part in planar joints and free joints.
Fig. 4.9 and table 4.1 (pp. 78-79) consider that the translations are expressed in successor frame (FS), but example 4.5 (p. 81) makes the case for having (translation) velocity variables expressed in FS but position variables expressed in parent frame (FP). This is to get better numerical stability while keeping simple motion subspace matrices `S`.

From the code of `Joint::pose`, planar joints are using the convention of Fig. 4.9, while free joints use that of example 4.5. I would push for having better coherency and change the convention of planar joint to the latter option: we would gain not only on the numerical behavior front, but also it is also, imho, more intuitive for the user to express the translation independently of the rotation.

In itself the change is drastic as it can break client code. In practice, it seems that the planar joint is not used much. 
I changed locally `Joint::pose` for planar joints, and it didn't break any test. It seems to me that this is the only place the code need to change, but I don't know the whole library well enough, so there might be other places to amend. Maybe @jorisv has some advice on that.

It would be better to take a decision before addressing #35 to avoid additional work.

Issue #35, opened 2017-05-26
Fix planar joints integration


gergondet:
Following #33, integration for planar joints needs to be looked at

gergondet:
So, I wanted to assign this to @aescande but I cannot for some reasons :crying_cat_face:

haudren:
Maybe because he is not a member of jrl-umi3218?

gergondet:
He is but he is a private member so that may be the issue...

aescande:
shhhh, don't tell everyone ;)

PR #66, opened 2019-07-09
Topic/integration

aescande:
This PR reworks the numerical integration for the non-trivial joints (planar, spherical and free):
* fix a sign error in the Magnus expansion and conditionally extend the expansion to 5 terms, depending on a user-defined precision,
* properly take into account that translation velocities are expressed in the successor frames (FS) by integrating the motion of this frame instead of considering it constant over the integration step,
* the error done during the integration is estimated and used to subdivide the integration step if necessary. Consequently, integration can be carried out on large intervals (tested up to 1 second),
* new tests are added, old ones are reworked
* detailed mathematical notes are added in the doc folder. Not yet linked to the doxygen.
* change name from `eulerIntegration` to `integration` or `numericalIntegration` to better reflect what's happening in the functions. Keep old names around but marked as deprecated.

This PR is not polished yet, but I'd like to have some general feedback first/while polishing.
First of all the tests. I'm currently performing 4 types:
 * comparing integration at constant velocity with analytical formula (for constant velocity it is possible to derive these formula, which is not the case for constant acceleration)
 * comparing the integration at constant acceleration with step dt to N integrations at constant velocity with step dt/N, changing the velocity between those integrations. N is a large number (10000 right now)
 * checking the self-consistency of the integration: one integration with step dt should yield the same result as N integrations with step dt/N. 
 * checking the consistency of the integration with the jacobian matrix in the following way: I take a point P on the body after the joint. Taking a constant velocity v for the joint, we should get the same result by
(i) integrating the joint position and computing the new position of P
(ii) integrating Jv, where J is the Jacobian matrix of the point.

The last test is an attempt to ensure that the integration is coherent with the library as a whole, because it could be possible to misunderstand the conventions adopted in RBDyn and write a wrong integration that could however pass the 3 first tests. The idea of using the Jacobian matrix is to get quantities in the world frame, where there is no specific convention, from whatever the original spaces are for the joint variables and velocities/accelerations. If anyone has a better idea, I'd be happy to hear it.
Something related is done in `EulerTestV2` in `AlgoTest`, but quantites are brought back in some variable space, which makes the test dependent on conventions, and makes it actually wrong for planar and free joints (although because of the very low accuracy required, the planar test is still passing - but I had to disable the one for free joint). I believe this test should be removed completely.

Another point is that the estimation of the error is adding some overhead. I didn't measure it for now, as it is up to a few dozens of scalar additions/multiplications, but it could be equivalent to the integration computation in itself. Does it seems ok? Or do I add a flag/versions without any precision check?

The PR solves #35, without addressing #65 (I could not get an integration coherent with the Jacobian matrix when changing the convention for the planar joint, and I don't get where I'm making a mistake).


aescande:
@gergondet : there does not seems to be a mechanism for deprecation in RBDyn yet. Do you have any guidelines on how to proceed?
Here are a few options I see:
* using `[[deprecated]]` (requires c++14)
* using some CMake facilities: https://stackoverflow.com/a/41225414/11611648
* writing some macro by hand (where?) as in https://stackoverflow.com/a/295229/11611648

aescande:
Baring feedbacks, I'm done with the code so far.
Code review can start :)

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+
+/** Compute the sum of the first terms of the Magnus expansion of \Omega such that
+ * q' = q*exp(\Omega) is the quaternion obtained after applying a constant acceleration
+ * rotation wD for a duration step, while starting with a velocity w
+ * The function computes the i first terms of the sum, such that ||O_i+1|| < absEps and
+ * ||O_i+1|| < relEps * ||O1||. It stops at the 5th term even if the conditions where not
+ * met, in which case the return bool is set to true as a warning.
+ */
+std::pair<Vector3d, bool> magnusExpansion(const Vector3d & w, const Vector3d & wD, double step,
```
vsamy: I think the return type is not intuitive, the function returns `false` if it worked and `true` if we don't know.
At the 5th step, there is no check whether or not the operation succeed and the return value is always `true` but could also be a `false`.

src/RBDyn/NumericalIntegration.h
```diff
@@ -0,0 +1,71 @@
...
+ * a warning, and qi is returned, otherwise perform the full computation.
+ */
+RBDYN_DLLAPI std::pair<Eigen::Quaterniond, bool> SO3Integration(
+                                                      const Eigen::Quaterniond & qi,
+                                                      const Eigen::Vector3d & wi,
+                                                      const Eigen::Vector3d & wD,
+                                                      double step,
+                                                      double relEps = 1e-12,
+                                                      double absEps = std::numeric_limits<double>::epsilon(),
+                                                      bool breakOnWarning = false);
```
vsamy: `breakOnWarning` should be renamed since no warning is emitted (not to the user/developer).

src/CMakeLists.txt
```diff
@@ -3,10 +3,10 @@
 #
 
 set(SOURCES MultiBodyGraph.cpp MultiBody.cpp MultiBodyConfig.cpp
-  FK.cpp FV.cpp FA.cpp Jacobian.cpp ID.cpp IK.cpp IS.cpp FD.cpp EulerIntegration.cpp
+  FK.cpp FV.cpp FA.cpp Jacobian.cpp ID.cpp IK.cpp IS.cpp FD.cpp NumericalIntegration.cpp
```
vsamy: I am not really use to deal with deprecation stuff but if `EulerIntegration.cpp` is deleted then the program just won't compile at all for old style and there is no need to mark it deprecated. Should'nt the file stay alive for few release iteration before it is definitively deleted?

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  return H;
+}
+
+} // namespace
+
+namespace rbd
+{
+
+std::pair<Quaterniond, bool> SO3Integration(const Quaterniond & qi,
```
vsamy: Is c++17 ok? If so, `std::optional` would suit exactly the behavior you want with the pairs.

src/RBDyn/NumericalIntegration.h
```diff
@@ -0,0 +1,71 @@
...
+ * Integrate a constant rotation acceleration.
+ * @param qi Initial orientation.
+ * @param wi Initial rotation speed.
+ * @param wD Constant acceleration.
+ * @param step Integration step.
+ * @param relEps Stopping criterion for the underlying Magnus expansion, relatively to the norm of its first term.
+ * @param absEps Absolute precision required on the underlying Magnus expansion.
+ * @param breakOnWarning If true, computation is stopped right after the underlying Magnus expansion when it returns
+ * a warning, and qi is returned, otherwise perform the full computation.
+ */
```
vsamy: Add an `@return` statement to explain the function output.

vsamy (commented):
The code seems to be clean. There is just some few changes.
Although, you might need to transform your `if (...) ...;` in 
```
if (...)
{
  ...
}
```

src/IS.cpp
```diff
@@ -54,7 +54,7 @@ void InverseStatics::setJacobianSize(const MultiBody & mb,
   {
     if(jacMomentsAndForces[i].cols() > nColsWanted)
     {
-      nColsWanted = jacMomentsAndForces[i].cols();
+      nColsWanted = static_cast<long>(jacMomentsAndForces[i].cols());
```
gergondet: very minor but it would make more sense to change `nColsWanted` type in the first place

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+ * rotation wD for a duration step, while starting with a velocity w
+ * The function computes the i first terms of the sum, such that ||O_i+1|| < absEps and
+ * ||O_i+1|| < relEps * ||O1||. It stops at the 5th term even if the conditions where not
+ * met, in which case the return bool is set to true as a warning.
+ */
+std::pair<Vector3d, bool> magnusExpansion(const Vector3d & w,
+                                          const Vector3d & wD,
+                                          double step,
+                                          double relEps,
+                                          double absEps)
```
gergondet: I think I'd rather have: (same goes for `SO3Integration`)
```cpp
Eigen::Quaterniond magnusExpansion(const Vector3d & w,
                                    const Vector3d & wD,
                                    double step,
                                    double relEps,
                                    double absEps,
                                    bool & failed)
```

this will make the code more readable and slightly more optimal

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+  double sqn5 = (sqndt4 * sqn1 + 8 * std::sqrt(sqndt4 * sqn1 * sqn2) + 16 * sqn2) * sqn1 * sqn2 / (840 * 840);
+
+  if(sqn5 < eps2)
+  {
+    return {O1 + O2 + O3 + O4, false};
+  }
+
+  Vector3d O5 = ((120 * sqn1 - 5 * sqndt4) * O3 - 24 * sqn2 * O1) / 5040;
+  return {O1 + O2 + O3 + O4 + O5, true};
+}
```
gergondet: Maybe we want a version that doesn't check bounds and just compute the 5 terms regardless (such as when you use it in `freeJointIntegration_`) but that has to be benchmarked....

In that version when you care about the boolean you are acutally computing O5 for nothing since you throw the result away

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+  Vector3d vh = vi + vD * halfNthStep;
+  Vector3d vf = vi + vD * nthStep;
+  Quaterniond qh = rbd::SO3Integration(H.first, wi, wD, halfNthStep).first;
+
+  H.second += nthStep * (4 * (qh * vh) + qf * vf);
+  H.first = qf;
+  H.second /= 6;
+
+  return H;
+}
```
gergondet: I had a short discussion with @aescande as this implementation didn't strictly match the equation in the note, in conclusion:
- we can change the inside of the loop to be marginally more precise
  - increment `wi` by `halfNthStep * wD` before computing `qe` (resp. change the increment at the end of the loop)
  - changing `qe` to `Quaterniond qe = rbd::SO3Integration(qh, wi, wD, halfNthStep).first;`
- we keep `qf` in the end although it would be technically more correct to compute `qf` from the final `qh`, the reason being precision is good enough and this avoids one SO3Integration in the most common use case (i.e. `n = 1`)

I suggest adding a note to "4.3 Implementation details" of the note to indicate that we take some shortcuts in the implementation since at this point we have made sure that `step` is sufficiently small to have good precision here. A comment here wouldn't hurt as well :-)

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+std::pair<Quaterniond, Vector3d> freeJointIntegration(const Quaterniond & qi,
+                                                      const Vector3d & xi,
+                                                      const Vector3d & wi,
+                                                      const Vector3d & vi,
+                                                      const Vector3d & wD,
+                                                      const Vector3d & vD,
+                                                      double step,
+                                                      double prec = 1e-10)
+{
+  auto qf = rbd::SO3Integration(qi, wi, wD, step, prec, prec);
```
gergondet: Should also break on failure

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+{
+  double erri = fourthDerivativeSquaredNormPlanar(vi, wi, vD, wD);
+  double errf = fourthDerivativeSquaredNormPlanar(vi + step * vD, wi + step * wD, vD, wD);
+  double errMax = std::sqrt(std::max(erri, errf));
+  int n = static_cast<int>(std::ceil(step / 2 * std::sqrt(std::sqrt(errMax * step / (180 * prec)))));
+  if(n == 0) n = 1;
+
+  double nthStep = step / n;
+  double halfNthStep = nthStep / 2;
+  double nthStep2 = nthStep * halfNthStep;
```
gergondet: Add a comment that this is the same as `nthStep * nThStep / 2`, this got me super confused ^^

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+  {
+    return {qi, true};
+  }
+  Vector3d O = mag.first / 2;
+  double n = O.norm();
+  double s = sva::sinc(n);
+  Quaterniond qexp(std::cos(n), s * O.x(), s * O.y(), s * O.z());
+
+  return {qi * qexp, mag.second};
+}
```
gergondet: I think this `breakOnWarning` parameter is very confusing (but very useful). I'd rather we have two functions:
- `SO3Integration` always returns the integration result regardless of magnus expansion result (that magnus expansion then doesn't perform as many checks)
- `SO3IntegrationIfGoodStep` (or a better name ^^) that checks the output of magnus expansion (and that magnus expansion doesn't have to bother with the fifth term computation)

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    /// @todo manage reverse joint
+    case Joint::Free:
+    {
+      Quaterniond qi(q[0], q[1], q[2], q[3]);
+      Map<Vector3d> xi(&q[4]);
+      Vector3d wi(alpha[0], alpha[1], alpha[2]);
+      Vector3d wD(alphaD[0], alphaD[1], alphaD[2]);
+      Vector3d vi(alpha[3], alpha[4], alpha[5]);
+      Vector3d vD(alphaD[3], alphaD[4], alphaD[5]);
```
gergondet: They could all be maps no?

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+      xi = H.second;
+      break;
+    }
+    /// @todo manage reverse joint
+    case Joint::Spherical:
+    {
+      Quaterniond qi(q[0], q[1], q[2], q[3]);
+      Vector3d wi(alpha[0], alpha[1], alpha[2]);
+      Vector3d wD(alphaD[0], alphaD[1], alphaD[2]);
```
gergondet: Could be Map also?

src/NumericalIntegration.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+      q[0] = qf.w();
+      q[1] = qf.x();
+      q[2] = qf.y();
+      q[3] = qf.z();
+      break;
+    }
+
+    case Joint::Fixed:
+    default:;
```
gergondet:
```cpp
default:
  break;
```

src/RBDyn/EulerIntegration.h
```diff
@@ -4,43 +4,46 @@
...
-#include <vector>
+#ifdef __GNUC__
+#  define DO_PRAGMA(x) _Pragma(#  x)
+#  define PRAGMA_WARNING(x) DO_PRAGMA(GCC warning #  x)
+#endif //__GNUC__
+#ifdef _MSC_VER
+#  define __STRINGIFY__(x) #  x
+#  define __TOSTRING__(x) __STRINGIFY__(x)
+#  define PRAGMA_WARNING(x) __pragma(message(__FILE__ "(" __TOSTRING__(__LINE__) ") : warning: " #  x))
+#endif
```
gergondet: Add `RBDYN` in all those macros name if they are going to be included :)

src/RBDyn/EulerIntegration.h
```diff
@@ -4,43 +4,46 @@
...
 
-// RBDyn
-#include <rbdyn/config.hh>
+PRAGMA_WARNING(EulerIntegration.h is a deprecated header.Please consider using NumericalIntegration.h instead.)
```
gergondet: -Please consider using
+Use

src/RBDyn/EulerIntegration.h
```diff
@@ -4,43 +4,46 @@
...
 
-/// safe version of @see eulerIntegration.
-RBDYN_DLLAPI void sEulerIntegration(const MultiBody & mb, MultiBodyConfig & mbc, double step);
+/// Old name for @see jointIntegration
+RBDYN_DEPRECATED void eulerJointIntegration(Joint::Type type,
+                                            const std::vector<double> & alpha,
+                                            const std::vector<double> & alphaD,
+                                            double step,
+                                            std::vector<double> & q,
+                                            double prec = 1e-10)
```
gergondet: Don't add the `prec` argument (applies to `eulerIntegration` and `sEulerIntegration` as well)

src/RBDyn/NumericalIntegration.h
```diff
@@ -0,0 +1,70 @@
...
+
+#pragma once
+
+// includes
+// std
+#include <vector>
+
+// RBDyn
+#include <rbdyn/config.hh>
+#include <rbdyn/deprecated.hh>
```
gergondet: I don't think this is needed here

gergondet (changes requested):
Sorry for the long time it took me to get to it. I don't have much issues with the PR, just small readability/efficiency concerns. I've rebased on the current master so you should make sure that you nuke your local branch before pushing back to it

I might be possible to integrate your latex note in the doxygen. I'll look into it

gergondet (approved):

PR #68, opened 2019-11-04
URDF and YAML parsers

BenjaminNavarro:
It's still work in progress but I wanted a place to discus about the issues I'm facing.

My goal is to integrate the URDF (from mc_rbdyn_urdf) and YAML (my own) parsers directly into the RBDyn repository and to provide facilities to convert between formats.

For the (very short) story, I made the YAML parser because I hate reading XML and even more writing it by hand. So if I need a model for a robot that doesn't have a URDF, I can now write it in YAML instead. 

The YAML parser has been extensively tested and provides the same functionalities as the URDF one.

Now, in order to convert between format, I need to be able to output URDF/YAML files from the RBDyn data structures. This leads to some issues:
1. Not all the data is available from `MultiBody`, `MultiBodyConfig` or `MultiBodyGraph`. For instance, joint limits are stored separately from the `Joint` class. This means that either we include them in the `Joint` class or we need to ask users to provide them separately. Modification of the `Joint` class can be made API backward compatible but not ABI backward compatible since new member variables must be added. The same goes for `visual`, `collision` and `collision_tf` from `ParserResult` that could be stored within the `Body` class instead.
2. The `parent`, `child` and `frame` components of the joints are not easy to get. So far, I couldn't find a proper way of extracting these information after the creation of the graph, but I'll keep looking.


src/parsers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+#
+# Copyright 2012-2019 CNRS-UM LIRMM, CNRS-AIST JRL
+#
+
+find_package(yaml-cpp REQUIRED)
```
gergondet: `add_project_dependency(yaml-cpp 0.5.0 REQUIRED)`

`add_project_dependency` is required to forward the dependency to other projects and you are using the modern API of yaml-cpp which requires `>= 0.5.0`

src/parsers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+          DESTINATION "${CONFIG_INSTALL_DIR}")
+  set(PACKAGE_EXTRA_MACROS "include(\${CMAKE_CURRENT_LIST_DIR}/FindRBDyn_TinyXML2.cmake)")
+else()
+  add_project_dependency(tinyxml2 REQUIRED)
+endif()
+
+set(SOURCES urdf.cpp yaml.cpp)
+set(HEADERS RBDyn/parsers/common.h RBDyn/parsers/urdf.h RBDyn/parsers/yaml.h)
+
+add_library(RBDynParsers SHARED ${SOURCES} ${HEADERS})
```
gergondet, line 25: `set_target_properties(RBDynParsers PROPERTIES EXPORT_NAME Parsers)`

I think it will be nicer to link to `RBDyn::Parsers` than to `RBDyn::RBDynParsers`

src/parsers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+target_include_directories(RBDynParsers PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include> $<INSTALL_INTERFACE:include>)
+target_link_libraries(RBDynParsers PUBLIC RBDyn tinyxml2::tinyxml2 yaml-cpp)
+set_target_properties(RBDynParsers PROPERTIES SOVERSION 1 VERSION 1.0.0)
```
gergondet: `set_target_properties(RBDynParsers PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR} VERSION ${PROJECT_VERSION_FULL})`

src/RBDyn/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  CoM.cpp Momentum.cpp ZMP.cpp IDIM.cpp VisServo.cpp Coriolis.cpp)
+set(HEADERS RBDyn/Body.h RBDyn/Joint.h RBDyn/MultiBodyGraph.h RBDyn/MultiBody.h RBDyn/MultiBodyConfig.h
+  RBDyn/FK.h RBDyn/FV.h RBDyn/FA.h RBDyn/Jacobian.h RBDyn/ID.h RBDyn/IK.h RBDyn/IS.h RBDyn/FD.h RBDyn/EulerIntegration.h RBDyn/CoM.h
+  RBDyn/Momentum.h RBDyn/ZMP.h RBDyn/IDIM.h RBDyn/VisServo.h RBDyn/util.hh RBDyn/util.hxx RBDyn/Coriolis.h)
+
+add_library(RBDyn SHARED ${SOURCES} ${HEADERS})
+target_include_directories(RBDyn PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include> $<INSTALL_INTERFACE:include>)
+target_link_libraries(RBDyn PUBLIC SpaceVecAlg::SpaceVecAlg)
+set_target_properties(RBDyn PROPERTIES COMPILE_FLAGS "-Drbdyn_EXPORTS")
+set_target_properties(RBDyn PROPERTIES SOVERSION 1 VERSION 1.1.0)
```
gergondet: `set_target_properties(RBDyn PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR} VERSION ${PROJECT_VERSION_FULL})`

src/parsers/RBDyn/parsers/api.h
```diff
@@ -0,0 +1,19 @@
...
+#else
+#  define RBDYNPARSERS_DLLIMPORT
+#  define RBDYNPARSERS_DLLEXPORT
+#endif
+
+#ifdef RBDYNPARSERS_BUILDING
+#  define RBDYNPARSERS_API RBDYNPARSERS_DLLEXPORT
+#else
+#  define RBDYNPARSERS_API RBDYNPARSERS_DLLIMPORT
+#endif
```
gergondet: Please use the same file as the one generated by jrl-cmakemodules (in your build directory: `include/rbdyn/config.hh`). It sets visibility flags for GCC

src/parsers/RBDyn/parsers/common.h
```diff
@@ -0,0 +1,112 @@
...
+	{
+		Superellipsoid() : size(Eigen::Vector3d::Zero()), epsilon1(1), epsilon2(1){
+		}
+		Eigen::Vector3d size;
+		double epsilon1,epsilon2;
+	};
+
+	enum Type { BOX, CYLINDER, SPHERE, MESH, SUPERELLIPSOID, UNKNOWN };
+	Type type;
+	boost::variant<Box, Cylinder, Mesh, Sphere, Superellipsoid> data;
```
gergondet: I'm surprised you don't have to include `<boost/variant.hpp>` here?

src/parsers/RBDyn/parsers/common.h
```diff
@@ -0,0 +1,112 @@
...
+	ParserResult& result() {
+		return res;
+	}
+
+	operator std::tuple<rbd::MultiBody&, rbd::MultiBodyConfig&, rbd::MultiBodyGraph&>() {
+		return std::tie(res.mb, res.mbc, res.mbg);
+	}
+
+	ParserResult res;
+};
```
gergondet, line 114: I don't think having a base class here is beneficial, free functions do the job as well...

In any case, it definitely doesn't have to be `virtual`

src/parsers/RBDyn/parsers/common.h
```diff
@@ -0,0 +1,112 @@
+#pragma once
```
gergondet: This file is missing its copyright notice

src/parsers/RBDyn/parsers/urdf.h
```diff
@@ -0,0 +1,66 @@
...
+ * (at your option) any later version.
+ *
+ * mc_rbdyn_urdf is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with mc_rbdyn_urdf.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
gergondet, line 3: Wrong header :D

src/parsers/RBDyn/parsers/urdf.h
```diff
@@ -0,0 +1,66 @@
...
+		const std::string & input,
+		ParserInput input_type = ParserInput::File,
+		bool fixed = true,
+		const std::vector<std::string> & filteredLinksIn = {},
+		bool transformInertia = true,
+		const std::string & baseLinkIn = "",
+		bool withVirtualLinks = true,
+		const std::string sphericalSuffix = "_spherical");
+
+ virtual ~RBDynFromURDF() = default;
```
gergondet: Doesn't have to be virtual or defaulted.

src/parsers/RBDyn/parsers/urdf.h
```diff
@@ -0,0 +1,66 @@
...
+Eigen::Vector3d attrToVector(const tinyxml2::XMLElement & dom, const std::string & attr, const Eigen::Vector3d & def = Eigen::Vector3d(0,0,0));
+
+Eigen::Matrix3d RPY(const double & r, const double & p, const double & y);
+
+rbd::Joint::Type rbdynFromUrdfJoint(const std::string & type);
+
+sva::PTransformd originFromTag(const tinyxml2::XMLElement & root, const std::string & tagName);
+sva::PTransformd originFromTag(const tinyxml2::XMLElement * dom);
+
+ParserResult rbdyn_from_urdf(const std::string & content, bool fixed = true, const std::vector<std::string> & filteredLinksIn = {}, bool transformInertia = true, const std::string & baseLinkIn = "", bool withVirtualLinks = true, const std::string sphericalSuffix = "_spherical");
```
gergondet: I'd rather have an overload of this function which accepts `ParserInput input_type` as a first argument than a class here.

src/parsers/RBDyn/parsers/yaml.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
gergondet: Missing copyright notice

src/parsers/urdf.cpp
```diff
@@ -0,0 +1,499 @@
...
+ * (at your option) any later version.
+ *
+ * mc_rbdyn_urdf is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with mc_rbdyn_urdf.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
gergondet, line 3: Wrong notice

src/parsers/urdf.cpp
```diff
@@ -0,0 +1,499 @@
...
+    ParserInput input_type,
+    bool fixed,
+    const std::vector<std::string> & filteredLinksIn,
+    bool transformInertia,
+    const std::string & baseLinkIn,
+    bool withVirtualLinks,
+    const std::string sphericalSuffix)
+{
+	if(input_type == rbd::ParserInput::File) {
+		std::ifstream file(input);
```
gergondet: You should check that the file opening was succesfull

src/parsers/urdf.cpp
```diff
@@ -0,0 +1,499 @@
...
+		                      fixed,
+		                      filteredLinksIn,
+		                      transformInertia,
+		                      baseLinkIn,
+		                      withVirtualLinks,
+		                      sphericalSuffix);
+	}
+}
+
+}
```
gergondet, line 504: This is not the most recent code from mc_rbdyn_urdf. It's missing all commits since at least: https://github.com/jrl-umi3218/mc_rbdyn_urdf/commit/703e0396db9c0aa15906c8992ed7ad8724d4eb30

src/parsers/yaml.cpp
```diff
@@ -0,0 +1,515 @@
...
+		{"fixed",rbd::Joint::Fixed}
+	};
+
+	YAML::Node config;
+	if(input_type == ParserInput::File) {
+		config = YAML::LoadFile(input);
+	}
+	else {
+		config = YAML::Load(input);
+	}
```
gergondet: This should catch parsing exceptions and file not found exceptions

tests/CMakeLists.txt
```diff
@@ -19,6 +19,17 @@ macro(addUnitTest name)
   endif()
 endmacro(addUnitTest)
 
+macro(addParserUnitTest name)
+  if(NOT ${DISABLE_TESTS} AND ${BUILD_PARSERS})
+    add_executable(${name} ${name}.cpp ${HEADERS})
+    target_link_libraries(${name} PRIVATE RBDynParsers Boost::unit_test_framework Boost::system Boost::filesystem Boost::disable_autolinking)
+    target_compile_definitions(${name} PRIVATE -DBOOST_TEST_DYN_LINK -DTESTS_DATA_DIR="${CMAKE_CURRENT_SOURCE_DIR}")
```
gergondet, line 26: I don't think you need this

tests/CMakeLists.txt
```diff
@@ -19,6 +19,17 @@ macro(addUnitTest name)
   endif()
 endmacro(addUnitTest)
 
+macro(addParserUnitTest name)
+  if(NOT ${DISABLE_TESTS} AND ${BUILD_PARSERS})
```
gergondet: use `BUILD_TESTING` instead of `NOT ${DISABLE_TESTS}`

tests/YAMLParserTest.cpp
```diff
@@ -0,0 +1,363 @@
...
+//  Root     j0   |   j1     j2
+//  ---- b0 ---- b1 ---- b2 ----b3
+//  Fixed    RevX   RevY    RevZ
+
+//  X
+//  ^
+//  |
+//   -- > Y
+
+std::string XYZSarmYaml(
```
gergondet, line 24: Use a raw-string as in the URDF test, this is much more readable

src/parsers/RBDyn/parsers/urdf.h
gergondet: While we're at it it could rename `from_urdf` so that we have the call `rbd::from_urdf` which looks nice to me :)

gergondet (changes requested):
Thanks for the contribution, that would be nice to drop mc_rbdyn_urdf in favor of a built-in parser

Apart from the requested changes, I would also like you to address:
- clang-format: just run `./.clang-format-fix.sh` in the repository to adapt your PR
- mc_rbdyn_urdf: update the parser code to the most recent version
- Python bindings: use the bindings of mc_rbdyn_urdf as a basis, that should mostly be a matter of renaming the namespace

gergondet:
> Now, in order to convert between format, I need to be able to output URDF/YAML files from the RBDyn data structures. This leads to some issues:

Sorry I forgot to answer about this...

Isn't the simplest solution to use the parser result to do the conversion rather than adding extra information to RBDyn classes?

gergondet:
Hi @BenjaminNavarro any updates on this?

BenjaminNavarro:
Sorry, not yet. I've been busy with other higher priority stuff lately. I'll try to get back to it this month

BenjaminNavarro:
Just to let you know that I'm finally back on this. I just rebased my branch on master and I'll start looking into @gergondet comments. 

I think this PR will only contain the parsers and then I'll make new ones for the YAML/URDF generators and for some core modifications we made in our own version of RBDyn

BenjaminNavarro:
Ok so the integration of the parsers is done and both tests pass.

In addition to `mc_rbdyn_urdf` and the YAML parser here are a few things that were added/improved:
 - Removed some UB and possible exceptions when accessing boost::variant
 - Support for the following shapes  in addition to the existing `Mesh`: `Box`, `Cylinder`, `Sphere` and `Superellipsoid`
 - Three free functions to load models: `from_urdf`, `from_yaml` and `from_file`. The two first ones can either load the description from a string or load it form a file. The last one automatically checks the file extension to chose which format to parse

I need to look at the Python bindings now.

The CI fails because of the missing dependencies. Should I also take care of this? I can also make the `BUILD_PARSERS` CMake option default to off.

gergondet:
> The CI fails because of the missing dependencies. Should I also take care of this? I can also make the BUILD_PARSERS CMake option default to off.

I have pushed a patch to take care of that. Hopefully it will go well :)

> Removed some UB and possible exceptions when accessing boost::variant

Out of curiosity, do you recall what were the specific changes?

> Support for the following shapes in addition to the existing Mesh: Box, Cylinder, Sphere and Superellipsoid

`Superllipsoid` is not part of the URDF format afaik, could you link some reference on the excepted parameters of the tag? (e.g. in doxygen)

> Three free functions to load models: from_urdf, from_yaml and from_file. The two first ones can either load the description from a string or load it form a file. The last one automatically checks the file extension to chose which format to parse

I think I prefer to have `from_urdf(content, ...)` and `from_urdf_file(file, ...)` rather than an overload with a tag to distinguish the two (and that will make the bindings much simpler to write as well)

It's getting late here so I'll have a closer look at it tomorrow but it looks like you haven't addressed some of the comments regarding the CMake scripts themselves. Otherwise, probably all good :)

src/parsers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+#
+# Copyright 2012-2019 CNRS-UM LIRMM, CNRS-AIST JRL
+#
+
+find_package(tinyxml2 REQUIRED)
```
gergondet: This causes the CI to fail on Linux as the tinyxml2 version shipped by Ubuntu does not have a proper CMake script

src/parsers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+else()
+  add_project_dependency(tinyxml2 REQUIRED)
+endif()
+
+set(SOURCES common.cpp urdf.cpp yaml.cpp)
+set(HEADERS RBDyn/parsers/common.h RBDyn/parsers/urdf.h RBDyn/parsers/yaml.h)
+
+add_library(RBDynParsers SHARED ${SOURCES} ${HEADERS})
+target_include_directories(RBDynParsers PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include> $<INSTALL_INTERFACE:include>)
+target_link_libraries(RBDynParsers PUBLIC RBDyn tinyxml2::tinyxml2 yaml-cpp)
```
gergondet: You need `Boost::boost` as well

src/parsers/CMakeLists.txt
BenjaminNavarro: Oops, don't know why this line reappeared since @arntanguy made a patch for this. I probably messed up my rebase. I'll remove it

src/parsers/CMakeLists.txt
BenjaminNavarro: Good catch!

BenjaminNavarro:
> > Removed some UB and possible exceptions when accessing boost::variant
> 
> Out of curiosity, do you recall what were the specific changes?

basically replacing things like:
```cpp
auto& mesh = boost::get<Geometry::Mesh>(v.geometry.data);
mesh.filename = meshDom->Attribute("filename");
v.geometry.data = mesh;
```

with:
```cpp
auto mesh = Geometry::Mesh();
mesh.filename = meshDom->Attribute("filename");
v.geometry.data = mesh;
```
Since it assumes that a `Geometry::Mesh` is the current active member of the variant, which might not be the case.

> > Support for the following shapes in addition to the existing Mesh: Box, Cylinder, Sphere and Superellipsoid
> 
> `Superllipsoid` is not part of the URDF format afaik, could you link some reference on the excepted parameters of the tag? (e.g. in doxygen)

Not it's not, but since it's part of sch-core and that it can be handy we added it some time ago. Yes, I'll link the wikipedia page which has a nice visual to understand what each parameter is doing.

> > Three free functions to load models: from_urdf, from_yaml and from_file. The two first ones can either load the description from a string or load it form a file. The last one automatically checks the file extension to chose which format to parse
> 
> I think I prefer to have `from_urdf(content, ...)` and `from_urdf_file(file, ...)` rather than an overload with a tag to distinguish the two (and that will make the bindings much simpler to write as well)

Ok no problem, I will do that

> It's getting late here so I'll have a closer look at it tomorrow but it looks like you haven't addressed some of the comments regarding the CMake scripts themselves. Otherwise, probably all good :)

I did, but I really messed up a rebase after that and lost the modifications... Sure, have a nice evening/night

BenjaminNavarro:
By the way, there is this fixme laying in the code:
```
// FIXME! Just like visual tags, there can be several collision tags!
```
Should I take care of it as well? Since fixing it would break the ABI that's probably a good time to address this. Unless there's no real need for it

arntanguy:
I wrote that comment when I added the parsing of the `<visual>` tags. Concerning collisions, I double checked why I left that `FIXME`, and indeed the [URDF specification](http://wiki.ros.org/urdf/XML/link#Elements) says:

> Note: multiple instances of \<collision\> tags can exist for the same link. The union of the geometry they define forms the collision representation of the link. 

Thus it would be better if the parser supported this. If you don't mind taking care of it that'd be great. Especially considering that I'm planning to add support for the URDF collision primitives in mc_rtc soon :)

gergondet (approved):

gergondet:
Hi,

I've made some small changes/cleanup on the PR and it's good for me atm.

Changes:
- Make Python bindings for the parsers a module of the rbdyn module (i.e. `rbdyn.parsers.from_urdf_file` (and add bindings for `from_file`)
- Use a different API macro for RBDynParsers (this avoids inconsistent DLL linkage issue in Windows)
- Rename `BUILD_PARSERS` to `BUILD_RBDYN_PARSERS`
- Completely remove `ParserResult::collision_tf` as this is redundant with the origin in the collection of `ParserResult::collision`
- Rewrote the URDF code to handle all types of visual/collision tags for both visual and collision (and also handle cylinder which was left out)
- Clean demangle function

I still have to make some changes for the debian packaging side.

I'm also thinking of moving all the parsers related code in to the `rbd::parsers` namespace, what do you think of it ? @arntanguy @BenjaminNavarro

BenjaminNavarro:
That's great, thanks a lot!

I don't see any problem in moving all parsers related code to `rbd::parsers`.

gergondet:
Ok, I've made the required changes on the packaging side and moved everything parsers related to the `rbd::parsers` namespace

Will merge when the CI passes and release `v1.2.0` shortly after

PR #72, opened 2020-04-30
Various improvements

BenjaminNavarro:
This PR brings two major improvements:
1. The ability to compute a Jacobian using any body as reference, without having to reconstruct a new graph. Thanks to @sonny-tarbouriech for the original implementation
2. Turn `Werror` back on. All warnings have been fixed and the CI passes on all platforms. The main sources of warnings were:
    1. Pessimistic moves: things like `return std::move(obj);` which disables copy elision
    2. signed/unsigned integer implicit conversions. All indexing is based on signed integers (partly to handle the -1 case) but STL containers use unsigned integers for size and indexing. All conversions are now explicit and performed using `static_cast`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,9 +19,6 @@ include(cmake/base.cmake)
 include(cmake/cython/cython.cmake)
 include(cmake/msvc-specific.cmake)
 
-# Disable -Werror on Unix for now.
-set(CXX_DISABLE_WERROR True)
-
 project(RBDyn CXX)
 
 option(BENCHMARKS "Generate benchmarks." OFF)
--- a/src/RBDyn/Jacobian.cpp
+++ b/src/RBDyn/Jacobian.cpp
@@ -30,10 +30,10 @@ Jacobian::Jacobian(const MultiBody & mb,
                    const Eigen::Vector3d & point)
 : jointsPath_(), point_(point), jac_(), jacDot_()
 {
-  body_index_ = mb.sBodyIndexByName(bodyName);
-  ref_index_ = mb.sBodyIndexByName(refBodyName);
+  bodyIndex_ = mb.sBodyIndexByName(bodyName);
+  refBodyIndex_ = mb.sBodyIndexByName(refBodyName);
 
-  int index = body_index_;
+  int index = bodyIndex_;
 
   int dof = 0;
   bool refindexFound = false;
@@ -43,7 +43,7 @@ Jacobian::Jacobian(const MultiBody & mb,
     dof += mb.joint(index).dof();
     reverseJoints_.insert(reverseJoints_.begin(), false);
 
-    if(index == ref_index_)
+    if(index == refBodyIndex_)
     {
       break;
     }
@@ -51,9 +51,9 @@ Jacobian::Jacobian(const MultiBody & mb,
   }
 
   // The two bodies don't belong to the same branch -> start from ref body to the root of the tree
-  if(index != ref_index_)
+  if(index != refBodyIndex_)
   {
-    index = ref_index_;
+    index = refBodyIndex_;
     int count = 0;
     // Add joints until reaching the common node
     do
@@ -72,10 +72,10 @@ Jacobian::Jacobian(const MultiBody & mb,
     {
       // Get to the common node
     }
-    dof -= std::accumulate(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx, 0,
+    dof -= std::accumulate(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx + 1, 0,
                            [&](int dofC, int idx) { return dofC + mb.joint(idx).dof(); });
-    jointsPath_.erase(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx);
-    reverseJoints_.erase(reverseJoints_.begin() + count, reverseJoints_.begin() + commonIdx);
+    jointsPath_.erase(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx + 1);
+    reverseJoints_.erase(reverseJoints_.begin() + count, reverseJoints_.begin() + commonIdx + 1);
   }
 
   jac_.resize(6, dof);
@@ -94,33 +94,47 @@ MultiBody Jacobian::subMultiBody(const MultiBody & mb) const
 
   for(int index = 0; index < static_cast<int>(jointsPath_.size()); ++index)
   {
-    int i = jointsPath_[static_cast<size_t>(index)];
+    int i = jointsPath_[index];
+
     // body info
     bodies.push_back(mb.body(i));
     parent.push_back(index - 1);
 
     // joint info
-    joints.push_back(mb.joint(i));
     succ.push_back(index);
     pred.push_back(index - 1);
-    Xt.push_back(mb.transform(i));
-  }
 
+    if (index == 0 )
+    {
+      if (reverseJoints_[index])
+        Xt.push_back(sva::PTransformd(Eigen::Vector3d(0., 0., 0.)));
+      else
+        Xt.push_back(mb.transform(i));
+    }
+    else
+    {
+      if (reverseJoints_[index-1])
+      {
+        if (reverseJoints_[index])
+          Xt.push_back(mb.transform(jointsPath_[index-1]).inv());
+        else
+          Xt.push_back(mb.transform(jointsPath_[index-1]).inv()*mb.transform(jointsPath_[index]));
+      }
+      else
+      {
+        Xt.push_back(mb.transform(i));
+      }
+    }
+    auto joint = mb.joint(i);
+    if (reverseJoints_[index])
+    {
+      auto fwd = joint.forward() ? false : true;
+      joint.forward(fwd);
+    }
+    joints.push_back(joint);
+  }
   return MultiBody(std::move(bodies), std::move(joints), std::move(pred), std::move(succ), std::move(parent),
                    std::move(Xt));
-}
-
-std::vector<int> Jacobian::dofPath(const MultiBody & mb) const
-{
-  std::vector<int> dof_path;
-  for(size_t i = 0; i < jointsPath_.size(); ++i)
-  {
-    for(int j = 0; j < mb.joint(jointsPath_[i]).dof(); ++j)
-    {
-      dof_path.push_back(jointsPath_[i]);
-    }
-  }
-  return dof_path;
 }
 
 /// private implementation of the Jacobian computation
@@ -132,8 +146,7 @@ static inline const Eigen::MatrixXd & jacobian_(const MultiBody & mb,
                                                 const Transform & Trans_0_p,
                                                 const std::vector<int> & jointsPath,
                                                 const std::vector<uint8_t> & reverseJoints,
-                                                Eigen::MatrixXd & jac,
-                                                int ref_index)
+                                                Eigen::MatrixXd & jac)
 {
   const std::vector<Joint> & joints = mb.joints();
   int curJ = 0;
@@ -155,14 +168,6 @@ static inline const Eigen::MatrixXd & jacobian_(const MultiBody & mb,
 
     curJ += joints[i].dof();
   }
-  // Change Jacobian base : root of the tree --> reference body
-  if(ref_index > 0)
-  {
-    auto dof_count = jac.cols();
-    jac.block(0, 0, 3, dof_count) = mbc.bodyPosW[ref_index].rotation() * jac.block(0, 0, 3, dof_count);
-    jac.block(3, 0, 3, dof_count) = mbc.bodyPosW[ref_index].rotation() * jac.block(3, 0, 3, dof_count);
-  }
-
   return jac;
 }
 
@@ -170,7 +175,15 @@ const Eigen::MatrixXd & Jacobian::jacobian(const MultiBody & mb,
                                            const MultiBodyConfig & mbc,
                                            const sva::PTransformd & X_0_p)
 {
-  return jacobian_(mb, mbc, X_0_p, jointsPath_, reverseJoints_, jac_, ref_index_);
+  jacobian_(mb, mbc, X_0_p, jointsPath_, reverseJoints_, jac_);
+  // Change Jacobian base : root of the tree --> reference body
+  if(refBodyIndex_ > 0)
+  {
+    auto dof_count = jac_.cols();
+    jac_.block(0, 0, 3, dof_count) = mbc.bodyPosW[refBodyIndex_].rotation() * jac_.block(0, 0, 3, dof_count);
+    jac_.block(3, 0, 3, dof_count) = mbc.bodyPosW[refBodyIndex_].rotation() * jac_.block(3, 0, 3, dof_count);
+  }
+  return jac_;
 }
 
 const Eigen::MatrixXd & Jacobian::jacobian(const MultiBody & mb, const MultiBodyConfig & mbc)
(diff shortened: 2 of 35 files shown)


CMakeLists.txt
```diff
@@ -19,9 +19,6 @@ include(cmake/base.cmake)
 include(cmake/cython/cython.cmake)
 include(cmake/msvc-specific.cmake)
 
-# Disable -Werror on Unix for now.
-set(CXX_DISABLE_WERROR True)
```
gergondet, line 23: I'd rather we do this target-based:

```cmake
option(RBDYN_TREAT_WARNING_AS_ERRORS "Treat warnings as errors" ON)
# Further down the line
if(RBDYN_TREAT_WARNING_AS_ERRORS)
  if(MSVC)
    target_compile_options(RBDyn PUBLIC $<BUILD_INTERFACE:/WX>)
  else()
    target_compile_options(RBDyn PUBLIC $<BUILD_INTERFACE:-Werror>)
  endif()
  # Alternatively, but I didn't try it
  target_compile_options(RBDyn PUBLIC $<BUILD_INTERFACE:$<IF:MSVC,/WX,-Werror>>)
endif()
```

This is cleaner cmake-wise and easier to disable on a case-by-case basis

src/RBDyn/RBDyn/Jacobian.h
```diff
@@ -49,6 +49,19 @@ class RBDYN_DLLAPI Jacobian
    */
   Jacobian(const MultiBody & mb, const std::string & bodyName, const Eigen::Vector3d & point = Eigen::Vector3d::Zero());
 
+  /**
+   * Create a jacobian from a reference body to the specified body.
+   * @param mb Multibody where bodyId is in.
```
gergondet: `bodyId` -> `bodyName`

Note: the original doxygen has the same issue, that would be nice to fix it as well

src/RBDyn/RBDyn/Jacobian.h
```diff
@@ -49,6 +49,19 @@ class RBDYN_DLLAPI Jacobian
...
+   * @param bodyName Specified body.
+   * @param refBodyName Reference body.
+   * @param point Point in the body exprimed in body coordinate.
```
gergondet: `exprimed` -> `expressed`

Note: the original doxygen has the same issue, that would be nice to fix it as well

src/RBDyn/RBDyn/Jacobian.h
```diff
@@ -307,6 +320,9 @@ class RBDYN_DLLAPI Jacobian
     return jointsPath_;
   }
 
+  /// @return The dof path vector from the root to the specified body.
+  std::vector<int> dofPath(const MultiBody & mb) const;
```
gergondet: What's the purpose of this function? I think the name is a little obscure, the doxygen could be clearer, e.g. `returns a flattened vector of dof indexes in that Jacobian path`

src/RBDyn/RBDyn/Jacobian.h
```diff
@@ -452,10 +468,14 @@ class RBDYN_DLLAPI Jacobian
 
 private:
   std::vector<int> jointsPath_;
+  std::vector<bool> reverseJoints_;
```
gergondet: `std::vector<bool>` has size over performance optimization so it's better to use `std::vector<uint8_t>` for this

src/RBDyn/RBDyn/MultiBody.h
```diff
@@ -73,13 +73,21 @@ class RBDYN_DLLAPI MultiBody
...
-    bodies_[num] = b;
+    bodies_[static_cast<std::size_t>(num)] = b;
+  }
+
+  /// @return Index of the root body
+  int rootBodyIndex() const
+  {
+    int index = bodyIndexByName(bodies_[0].name());
+    while(parent(index) != -1) index = parent(index);
+    return index;
```
gergondet: I don't think this function makes sense. It should always return 0

src/parsers/RBDyn/parsers/common.h
```diff
@@ -23,7 +23,7 @@ namespace rbd
 namespace parsers
 {
 
-enum class RBDYN_PARSERS_DLLAPI ParserInput
+enum class ParserInput
```
gergondet, line 26: I don't think this should be removed

src/RBDyn/Jacobian.cpp
```diff
@@ -21,13 +21,17 @@ Jacobian::Jacobian() {}
 Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eigen::Vector3d & point)
 : jointsPath_(), point_(point), jac_(), jacDot_()
```
gergondet: `: Jacobian(mb, bodyName, mb.body(0).name(), point)`

src/RBDyn/Jacobian.cpp
```diff
@@ -36,6 +40,63 @@ Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eig
...
+  while(!refindexFound && index != -1)
+  {
+    jointsPath_.insert(jointsPath_.begin(), index);
+    dof += mb.joint(index).dof();
+    reverseJoints_.insert(reverseJoints_.begin(), false);
+
+    if(index == ref_index_)
+      refindexFound = true;
+    else
+      index = mb.parent(index);
```
gergondet: please remove the boolean and rewrite as:
```cpp
if(index == ref_index_)
{
  break;
}
index = mb.parent(index);
```

src/RBDyn/Jacobian.cpp
```diff
@@ -36,6 +40,63 @@ Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eig
...
+    // Add joints until reaching the common node
+    do
+    {
+      jointsPath_.insert(jointsPath_.begin() + count, index);
+      dof += mb.joint(index).dof();
+      reverseJoints_.insert(reverseJoints_.begin(), true);
+
+      index = mb.parent(index);
+      count++;
+    } while(std::find(jointsPath_.begin(), jointsPath_.end(), index) == jointsPath_.end());
```
gergondet: you can change `jointsPath_.begin()` to `jointsPath_.begin() + count`

src/RBDyn/Jacobian.cpp
```diff
@@ -36,6 +40,63 @@ Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eig
...
+    // Delete common joints previously added
+    bool commonJoints = true;
+    while(commonJoints)
+    {
+      const auto joint_index = jointsPath_[static_cast<size_t>(count)];
+      if(joint_index == index) commonJoints = false;
+      dof -= mb.joint(joint_index).dof();
+      jointsPath_.erase(jointsPath_.begin() + count);
+      reverseJoints_.erase(reverseJoints_.begin() + count);
+    }
```
gergondet, line 73: I think this can be rewritten in a slightly more compact way:

```cpp
int commonIdx = count; 
while(jointsPaths_[++commonIdx] != index); // Get to the common node
dof -= std::accumulate(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx, 0, [&](int dofC, int idx) { return dofC + mb.joint(idx).dof(); });
jointsPath_.erase(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx);
reverseJoints_erase(jointsPath_.begin() + count, jointsPath_.begin() + commonIdx);
```

src/RBDyn/Jacobian.cpp
```diff
@@ -64,6 +125,16 @@ MultiBody Jacobian::subMultiBody(const MultiBody & mb) const
                    std::move(Xt));
 }
 
+std::vector<int> Jacobian::dofPath(const MultiBody & mb) const
+{
+  std::vector<int> dof_path;
+  for(size_t i = 0; i < jointsPath_.size(); ++i)
+  {
+    for(int j = 0; j < mb.joint(jointsPath_[i]).dof(); ++j) dof_path.push_back(jointsPath_[i]);
```
gergondet: rewrite as:
```cpp
for(...)
{
...
}
```

src/RBDyn/Jacobian.cpp
```diff
@@ -81,17 +154,33 @@ static inline const Eigen::MatrixXd & jacobian_(const MultiBody & mb,
...
 
     curJ += joints[i].dof();
   }
+  // Change Jacobian base : root of the tree --> reference body
+  if(ref_index != -1)
+  {
+    auto dof_count = jac.cols();
+    const auto joint_index = static_cast<size_t>(jointsPath[0]);
+    jac.block(0, 0, 3, dof_count) = mbc.bodyPosW[joint_index].rotation() * jac.block(0, 0, 3, dof_count);
+    jac.block(3, 0, 3, dof_count) = mbc.bodyPosW[joint_index].rotation() * jac.block(3, 0, 3, dof_count);
```
gergondet: I think it should be `ref_index` here and not `joint_index` as this is mixing joint and body indexes.

Also by construction the root of the tree is the reference body so I'm not even sure you need to apply this transformation.

src/RBDyn/Jacobian.cpp
```diff
@@ -81,17 +154,33 @@ static inline const Eigen::MatrixXd & jacobian_(const MultiBody & mb,
...
   {
-    int i = jointsPath[index];
+    const auto i = static_cast<size_t>(jointsPath[index]);
 
     sva::PTransformd X_i_N = X_0_p * mbc.bodyPosW[i].inv();
 
     for(int dof = 0; dof < joints[i].dof(); ++dof)
     {
-      jac.col(curJ + dof).noalias() = (X_i_N * (sva::MotionVecd(mbc.motionSubspace[i].col(dof)))).vector();
+      if(!reverseJoints[index])
```
gergondet: This check should go outside the loop, e.g.

```cpp
double sgn = reverseJoints[index] ? -1 : 1;
for(int dof = 0; dof < joints[i].dof(); ++dof)
{
  jac.col(curJ + dof).noalias() = (X_i_N * (sva::MotionVecd(sgn * mbc.motionSubspace[i].col(dof)))).vector();
}
```

gergondet (changes requested):
Thanks for taking the time of fixing all those pesky conversion warnings :)

Could you also add a unit test to check the functionality of taking a jacobian from a different reference body? Preferably with a somewhat complex case for this:

```
       D
       |
       C
     /   \
   B       E
  /         \
A            F
```

For example with such a tree, compute the jacobian for A using F as a reference body

src/RBDyn/RBDyn/Jacobian.h
BenjaminNavarro: I'm not sure, maybe @sonny-tarbouriech can shed some light on this one

src/RBDyn/RBDyn/MultiBody.h
BenjaminNavarro: Again, maybe @sonny-tarbouriech had a reason for this. I'll ask him to come and comment on this

src/parsers/RBDyn/parsers/common.h
BenjaminNavarro: I had compilation issues in CI because of this. It seems that not all compilers support this well and from what I gathered, it is not necessary (see https://stackoverflow.com/a/40402637/6458469 and https://stackoverflow.com/a/18641809/6458469)

src/RBDyn/Jacobian.cpp
BenjaminNavarro: If I'm not mistaken, the transformation is needed because you can now choose to use any body of the tree as reference

BenjaminNavarro:
I'll see how to add a unit test for that, even though it might be a bit difficult to test properly.

I commented when necessary and will implement your suggestions otherwise

CMakeLists.txt
BenjaminNavarro: So I create one option per target? One for RBDyn, one for the parsers and one for the tests/benchmarks?

src/RBDyn/RBDyn/Jacobian.h
sonny-tarbouriech: I agree that the purpose of this function is not that clear. The idea is to return a vector of joint indexes in the Jacobian path based on their number of dofs. It allows discarding fixed joints and taking into account multi-dimensional joints by adding the index multiple times in the returned vector. Don't know how to rename the function to be more explicit. Any suggestion?

src/RBDyn/RBDyn/MultiBody.h
sonny-tarbouriech: I was not sure that it would always return 0. But if this is the case we can remove this function.

src/RBDyn/Jacobian.cpp
sonny-tarbouriech: @gergondet you're right I think I mixed up joint and body indexed, so ```ref_index``` should be used instead. As @BenjaminNavarro mentioned, with the new feature the reference body is no longer necessarily the root of the tree but can be any body in the chain.

CMakeLists.txt
gergondet: No. As long as it is enabled for RBDyn it will propagate for the targets that link with RBDyn (only in the current build, hence the `$<BUILD_INTERFACE>` expression)

src/RBDyn/RBDyn/Jacobian.h
gergondet: I don't have a suggestion in mind. What do you use this for?

src/RBDyn/RBDyn/MultiBody.h
gergondet: Actually it is guaranteed to be zero if the `MultiBody` was built from a `MultiBodyGraph` but you could construct a `MultiBody` by hand that does not follow that convention. However, all algorithms assume that body 0 is the root of the tree... I think rather than having this function we should check that `parent(0) == -1` when constructing the `MultiBody`

src/parsers/RBDyn/parsers/common.h
gergondet: Oh ok, it seems Windows is happy to build without it so that's probably fine :)

src/RBDyn/Jacobian.cpp
gergondet: After thinking on it, I think the root of the tree is irrelevant here as the root disappears (only the relative transformation between a joint in the chain and the jacobian's body matters) but I might be missing something (perhaps it will be clearer with tests)

sonny-tarbouriech:
I did some modifications in the Jacobian class to make the new feature (the ability to compute a Jacobian using any body as reference) compatible with the original version. 

I also modified the `Jacobian::subMultiBody` function in order to return a MultiBody going from the reference frame to the body frame by adapting the transforms between the bodies and the joint directions when the kinematic path is reversed. This way, computing FK, FV and FA for this MB  directly gives the poses, velocities and accelerations of bodies expressed wrt the reference frame.

I implemented a unit test with a 2D dual-arm robot to check that the Jacobian computation and the velocities from the subMultiBody give consistent results. 

There is still one fail in CI with Windows but I don't know why.

@gergondet could you tell me if the changes are OK? Do you know what's the problem with the CI?

aescande:
> There is still one fail in CI with Windows but I don't know why.

You have a bug due to a mismatch between l314 and l319 of JacobianTest.cpp: you're creating a dual arm with floating base, but you associate an empty configuration to it. This ends up crashing in  `forwardKinematics->Joint::pose<double>->QuatToE<double>` when the conversion tries to access non-existing elements.

I'm surprised only windows STL catches that

sonny-tarbouriech:
> > There is still one fail in CI with Windows but I don't know why.
> 
> You have a bug due to a mismatch between l314 and l319 of JacobianTest.cpp: you're creating a dual arm with floating base, but you associate an empty configuration to it. This ends up crashing in `forwardKinematics->Joint::pose<double>->QuatToE<double>` when the conversion tries to access non-existing elements.
> 
> I'm surprised only windows STL catches that

Thanks Adrien, I wanted to see if the test would pass with a floating-based robot but forgot to update the joint configuration.

gergondet:
Sorry for the very long delay on this. I've rebased on master and merged, thanks for the work @BenjaminNavarro @sonny-tarbouriech

PR #73, opened 2020-05-05
[parsers] URDF and YAML generation from ParserResult

BenjaminNavarro:
Unless I'm missing something, the generated URDF/YAML models must be later read with `transformInertia = false` because there is no way to recover the original rotation from the available data. I don't know if it is a big issue or if we can find a workaround for this


src/parsers/CMakeLists.txt
```diff
@@ -39,3 +39,16 @@ install(
...
 install(FILES ${HEADERS} DESTINATION ${INCLUDE_INSTALL_DESTINATION}/parsers)
+
+
+add_executable(urdf_yaml_converter urdf_yaml_conv.cpp)
+target_link_libraries(urdf_yaml_converter RBDyn::Parsers)
+set_target_properties(urdf_yaml_converter PROPERTIES COMPILE_FLAGS "-DRBDYN_PARSERS_EXPORTS")
+
+install(
+    TARGETS urdf_yaml_converter
+    EXPORT "${TARGETS_EXPORT_NAME}"
```
gergondet: I don't think it's useful to export an executable

src/parsers/CMakeLists.txt
```diff
@@ -39,3 +39,16 @@ install(
     RUNTIME DESTINATION "bin"
 )
 install(FILES ${HEADERS} DESTINATION ${INCLUDE_INSTALL_DESTINATION}/parsers)
+
+
+add_executable(urdf_yaml_converter urdf_yaml_conv.cpp)
+target_link_libraries(urdf_yaml_converter RBDyn::Parsers)
+set_target_properties(urdf_yaml_converter PROPERTIES COMPILE_FLAGS "-DRBDYN_PARSERS_EXPORTS")
```
gergondet: This is consuming the parsers library so we don't want that

src/parsers/CMakeLists.txt
```diff
@@ -39,3 +39,16 @@ install(
     RUNTIME DESTINATION "bin"
 )
 install(FILES ${HEADERS} DESTINATION ${INCLUDE_INSTALL_DESTINATION}/parsers)
+
+
+add_executable(urdf_yaml_converter urdf_yaml_conv.cpp)
+target_link_libraries(urdf_yaml_converter RBDyn::Parsers)
```
gergondet: missing `PUBLIC`

src/parsers/urdf.cpp
```diff
@@ -93,6 +93,10 @@ Eigen::Vector3d attrToVector(const tinyxml2::XMLElement & dom, const std::string
 
 Eigen::Matrix3d RPY(const double & r, const double & p, const double & y)
 {
+  // return (Eigen::AngleAxisd{r, Eigen::Vector3d::UnitX()} * Eigen::AngleAxisd{p, Eigen::Vector3d::UnitY()}
+  //         * Eigen::AngleAxisd{y, Eigen::Vector3d::UnitZ()})
+  //     .toRotationMatrix();
+
```
gergondet: Please remove this

src/parsers/urdf_yaml_conv.cpp
```diff
@@ -0,0 +1,135 @@
...
+    else
+    {
+      std::cerr << "Failed to open " << file_path << " for writing" << std::endl;
+      std::exit(-1);
+    }
+  }
+  else
+  {
+    std::cout << conversion_result << std::endl;
+  }
```
gergondet, line 144: `return 0;`

src/parsers/urdf_yaml_conv.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  if(argc > 3)
+  {
+    print_usage();
+    std::exit(-1);
+  }
+
+  const auto stdin_content = getContent(std::cin);
+  const auto file_input = stdin_content.size() == 0;
+  const auto file_output = (file_input && argc == 3) || (!file_input && argc == 2);
```
gergondet, line 61: I'm a little confused here about how it works with the standard input?

gergondet (changes requested):
Looks ok (I haven't looked very carefully at the conversion code itself but I'm assuming the tests cover it well)

I'd like to see a few changes:
1. Factor the testing code for comparing `ParserResult` instances and generating the examples since they are basically the same between each tests
2. I don't like that you need to have `transformInertia = false` with the generated files, you can get an inertia that will transform correctly by calling `sva::inertiaToOrigin(inertia, -mass, com, Eigen::Matrix3d::Identity())`. The only information that is definitely lost is the inertial frame orientation that could be in the original URDF.

src/parsers/CMakeLists.txt
BenjaminNavarro: Sorry, dumb copy/paste...

src/parsers/CMakeLists.txt
BenjaminNavarro: Why? It's a utility to convert between formats, it needs the parsers library for that.

src/parsers/CMakeLists.txt
BenjaminNavarro: Is it really useful for executables?

src/parsers/urdf_yaml_conv.cpp
BenjaminNavarro: Why? It is not necessary and the default behavior

src/parsers/CMakeLists.txt
gergondet: Adding `-DRBDYN_PARSERS_EXPORTS` will change the API declaration to `__declspec(dllexport)` but we want `__declspec(dllimport)` since we are consumers of the library

src/parsers/CMakeLists.txt
gergondet: No but this is mainly for consistency sake :)

src/parsers/urdf_yaml_conv.cpp
BenjaminNavarro: `getContent` tries to read everything available on the given input stream, either `std::cin` or a file. So if `stdin_content` is empty, it means that there's nothing to read from `stdin` and that the input must be given as a file. Then the output could be either `stdout` or a file and that depends on both the type of input and number of arguments.

So in the end the program can be called like this:
1. `urdf_yaml_converter` : reads from `stdin` and writes to `stdout`
2. `urdf_yaml_converter out.ext` : reads from `stdin` and writes to `out.ext`
3. `urdf_yaml_converter in.ext` : reads from `in.ext` and writes to `stdout`
4. `urdf_yaml_converter in.ext out.ext` : reads from `in.ext` and writes to `out.ext`

BenjaminNavarro:
>     1. Factor the testing code for comparing `ParserResult` instances and generating the examples since they are basically the same between each tests
Ok no problem

>     2. I don't like that you need to have `transformInertia = false` with the generated files, you can get an inertia that will transform correctly by calling `sva::inertiaToOrigin(inertia, -mass, com, Eigen::Matrix3d::Identity())`. The only information that is definitely lost is the inertial frame orientation that could be in the original URDF.
I don't like it either so I'll try what you suggest and see if it works properly

src/parsers/urdf_yaml_conv.cpp
gergondet: Thanks for the clarification, but how do you distinguish case 2 and 3?

src/parsers/urdf_yaml_conv.cpp
gergondet: Thanks, I learned something, I didn't know `return 0;` was implicit for the `main` function. I still prefer to have it explicit out of habit :)

src/parsers/urdf_yaml_conv.cpp
BenjaminNavarro: It depends if there is data available on `stdin` or not.

gergondet: So, typically for 2
```bash
$ cat in.urdf | urdf_yaml_converter out.yaml
```
And for 3:
```bash
$ urdf_yaml_converter in.urdf
```
right?

BenjaminNavarro:
Ok, so the parser tests refactoring is done and `transformInertia=false` is no longer necessary! Let's see if the CI is happy

src/parsers/urdf_yaml_conv.cpp
BenjaminNavarro: Exactly

BenjaminNavarro: Just realized that 2. is not explained in the usage, I'll update it

src/parsers/CMakeLists.txt
BenjaminNavarro: I changed it to `-DRBDYN_PARSERS_DLLIMPORT`

gergondet (approved):

gergondet:
Thanks for your contribution @BenjaminNavarro

PR #74, opened 2020-07-24
add mbc member 'bodyAccW'

ndehio:
It would be convenient to access bodyAccW() in MultiBodyConfig.
I added the corresponding member, but it still need to be computed correctly. Is this done in IS.cpp and ID.cpp? And how?


gergondet:
This can be done in `FA.cpp`, the same way it's done in [`FV.cpp` for updating `bodyVelW` from `bodyVelB`](https://github.com/jrl-umi3218/RBDyn/blob/master/src/RBDyn/FV.cpp#L35). This can also be added in `ID.cpp`

PR #75, opened 2020-08-21
Compute the whole-body-inertia and the average-angular-velocity. 

wyqsnddd:
As I discussed in the [mc_rtc issue 245](https://gite.lirmm.fr/multi-contact/mc_rtc/-/issues/245), these commit provide additional computations.


src/RBDyn/Momentum.cpp
```diff
@@ -13,6 +13,41 @@
...
+    cmc += (mbc.bodyPosW[i] * X_com_0).transMul(hi).vector();
+
+    // sum: X^T_com_i*I_i*X_com_i
+    // X_com_i = X_i_0 * X_com_0
+    ci += ((mbc.bodyPosW[i] * X_com_0).matrix().transpose()) * (bodies[i].inertia().matrix())
+          * (mbc.bodyPosW[i] * X_com_0).matrix();
+  }
+
+  // Compute the average velocity: inertia.inverse()*momentum
+  av = ci.inverse() * cmc;
```
gergondet: Eigen does not perform especially well on 6x6 matrix so you are better of using a solve method to get the result here, e.g.:  (see [this page](https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html) for more details)
```c++
av = ci.partialPivLu().solve(cmc);
```

(from a quick bench on my side this is around 2 times faster than the inverse product on a 6x6 matrix

If the `ci` matrix is postive definite you can use LLT to solve the problem which is even more efficient

src/RBDyn/RBDyn/Momentum.h
```diff
@@ -21,6 +21,23 @@ struct Block;
 using Blocks = std::vector<Block>;
 class Jacobian;
 
+/**
+ * Compute the centroidal momentum at the CoM frame
```
gergondet: This comment is not correct

src/RBDyn/RBDyn/Momentum.h
```diff
@@ -21,6 +21,23 @@ struct Block;
...
+ * as describe in [Orin and Gosawami 2008].
+ * @param mb MultiBody used has model.
+ * @param mbc Use bodyPosW, bodyVelB.
+ * @param com CoM position.
+ * @param ci Centroidal inertia at the Centroidal frame.
+ * @param cm centroidal momentum at the Centroidal frame.
+ * @param av Average velocity is inverse(centroidalInertia)*centroidalMomentum
```
gergondet, line 76: I think we should introduce an overload for computing this:
```c++
// Skip average velocity computation
computeCentroidalInertia(mb, mbc, com, ci, cm);
// Do the extra computation for av
computeCentroidalInertia(mb, mbc, com, ci, cm, av);
```

src/RBDyn/Momentum.cpp
```diff
@@ -13,6 +13,41 @@
...
+                                         Eigen::Matrix6d & ci,
+                                         Eigen::Vector6d & cm,
+                                         Eigen::Vector6d & av)
+{
+
+  using namespace Eigen;
+
+  const std::vector<Body> & bodies = mb.bodies();
+  // cm Candidate
+  Vector6d cmc = Vector6d::Zero();
```
gergondet: What is the point of introducing cmc?

gergondet (changes requested):
This looks great, thanks!

Can you add some tests to check the computations are ok?

src/RBDyn/Momentum.cpp
wyqsnddd: I removed it.

src/RBDyn/RBDyn/Momentum.h
wyqsnddd: Good idea.

src/RBDyn/RBDyn/Momentum.h
wyqsnddd: Should be fixed now

src/RBDyn/Momentum.cpp
wyqsnddd: Indeed `ci` is positive definite. I replaced with `LLT`. Thanks for the tip!

wyqsnddd: It is a good idea to add some `test-case`. But I am wondering that do we have ground truth (known centroidal inertia) to test against?

aescande: > 
> 
> Eigen does not perform especially well on 6x6 matrix so you are better of using a solve method to get the result here, e.g.: (see [this page](https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html) for more details)
> 
> ```c++
> av = ci.partialPivLu().solve(cmc);
> ```
> 
> (from a quick bench on my side this is around 2 times faster than the inverse product on a 6x6 matrix
> 
> If the `ci` matrix is postive definite you can use LLT to solve the problem which is even more efficient

Just my grain of salt/preaching here, to add on Pierre's comment. It's not a problem of Eigen here. As a rule of thumb, you should never compute explicitly the inverse of a matrix, but rely on matrix decomposition and system solving, which is much more efficient.

If you don't need `cm` afterward, you can even go for
`ci.llt().solveInPlace(cm);`
which will be slightly faster (the value of `av` will now be found in `cm`).

wyqsnddd: @aescande Hi Adrien, thanks a lot for the discussion and the rule of thumb about the matrix inversion. I changed the new APIs to: 

```cpp

/**
 * Compute the centroidal inertia (ci), centroidal momentum (cm),
 * and the average velocity (av = ci^-1*cm)
 * at the CoM frame as describe in [Orin and Gosawami 2008].
 * @param mb MultiBody used has model.
 * @param mbc Use bodyPosW, bodyVelB.
 * @param com CoM position.
 * @param ci Centroidal inertia at the Centroidal frame.
 * @param cm centroidal momentum at the Centroidal frame.
 * @param av Average velocity is: av = ci.inverse()*cm
 */
RBDYN_DLLAPI void computeCentroidalInertia(const MultiBody & mb,
                                           const MultiBodyConfig & mbc,
                                           const Eigen::Vector3d & com,
                                           Eigen::Matrix6d & ci,
                                           Eigen::Vector6d & cm,
                                           Eigen::Vector6d & av);
/**
 * Compute the centroidal inertia (ci), centroidal momentum (cm),
 * and the average velocity (av = ci^-1*cm)
 * at the CoM frame as describe in [Orin and Gosawami 2008].
 * @param mb MultiBody used has model.
 * @param mbc Use bodyPosW, bodyVelB.
 * @param com CoM position.
 * @param ci Centroidal inertia at the Centroidal frame.
 * @param av Average velocity is inverse(centroidalInertia)*centroidalMomentum
 *
 * If the Centroidal Momentum is needed, we can easily compute it as: cm = ci*av.
 */
RBDYN_DLLAPI void computeCentroidalInertia(const MultiBody & mb,
                                           const MultiBodyConfig & mbc,
                                           const Eigen::Vector3d & com,
                                           Eigen::Matrix6d & ci,
                                           Eigen::Vector6d & av);


``` 

In the second case, when `cm` is not needed, `av` takes its place.

wyqsnddd:
In my other repo, I made a [quick assertion](https://github.com/wyqsnddd/mc_impact_predictor/blob/fidynamics/src/ImpactDynamicsModelInterface.cpp#L58) to make sure the `COM velocity` is equal to the `Average Linear Velocity`. Thus I think the computation should be correct.

gergondet:
> In my other repo, I made a quick assertion to make sure the COM velocity is equal to the Average Linear Velocity. Thus I think the computation should be correct.

Hi Yuquan,

I added a [test](https://github.com/jrl-umi3218/RBDyn/pull/75/commits/456756dda5668c4fb737384ead413cffc25e4abb) that test that but the test is failing. Could you have a look at it?

I also rebased your branch on master and fixed warnings on clang to get the build to run correctly, I had to force push to your branch so please be careful when you pull on your side

wyqsnddd:
> > In my other repo, I made a quick assertion to make sure the COM velocity is equal to the Average Linear Velocity. Thus I think the computation should be correct.
> 
> Hi Yuquan,
> 
> I added a [test](https://github.com/jrl-umi3218/RBDyn/pull/75/commits/456756dda5668c4fb737384ead413cffc25e4abb) that test that but the test is failing. Could you have a look at it?
> 
> I also rebased your branch on master and fixed warnings on clang to get the build to run correctly, I had to force push to your branch so please be careful when you pull on your side

Yes, sure! I'll look into the test.

wyqsnddd:
@gergondet Hi Pierre,

I am revisiting this issue again. I found something fishy with the `comVelocity`. Basically, I failed the following: 

```cpp
    BOOST_CHECK_SMALL((av.tail(3) - comVelocity).norm(), TOL);
```

but passed: 

```cpp
    BOOST_CHECK_SMALL((ci*av - cm.vector()).norm(), TOL);
```

I computed the `centroidal inertia` ci [here](https://github.com/wyqsnddd/RBDyn/blob/whole-body-inertia/src/RBDyn/Momentum.cpp#L101), which is mathematically correct according to the following equations. The computation of `av` should be fine too.

Reference equations (I can provide further reference on this point if necessary): 

- Inertia transform from frame $F_i$ to $F_c$: $Adg^\top_{i,c} I_i Adg_{i,c}$, where $I_i$ denotes the inertia in frame $F_i$. 
- Force / Momentum transform from $F_i$ to $F_c$: $Adg^\top_{i,c} h_i$, where $h_i$ denotes the momentum in frame $F_i$. 


Further, the linear momentum should be equal to the product of the COM velocity and the mass. But RBDyn can not pass such a test and the difference is huge if I do the following: 
```cpp
cmm.computeMatrix(mb, mbc, com);

    Eigen::MatrixXd cmmMatrix = cmm.matrix();

    Eigen::VectorXd alpha(mb.nrDof());
    double mass = 6.0;
    std::cout<<" Momemtum diff is: "<<((cmmMatrix * alpha).tail(3) - mass*comVelocity).transpose();
```
In  the above example, I found the mass from other computations.

gergondet:
Hi Yuquan,

```cpp
cmm.computeMatrix(mb, mbc, com);

    Eigen::MatrixXd cmmMatrix = cmm.matrix();

    Eigen::VectorXd alpha(mb.nrDof());
    double mass = 6.0;
    std::cout<<" Momemtum diff is: "<<((cmmMatrix * alpha).tail(3) - mass*comVelocity).transpose();
```

This sample as such is not correct, your alpha vector is a vector of size `mb.nrDof()` of unitialized values not the velocity vector.

wyqsnddd:
Hi Pierre, 

Sorry for the mistake. It is a redundant comment. Now I have removed it. 

Could you please run the example? The issue remains there.

wyqsnddd:
Hi Pierre, 

This is a kind reminder of the open issue. Maybe we can look at it together if you want. 


Yuquan

PR #76, opened 2020-11-04
[parsers] Load material from URDF

gergondet:
This loads the material tag information from URDF files

@BenjaminNavarro do you support these tags in the yaml format? If so I will add it to the YAML parser as well


BenjaminNavarro:
No we don't use the tag, at least right now. But it might still be a good idea to implement it in order to keep feature parity between the two parsers and keep the URDF<->YAML converter functional

gergondet:
Thanks for the feedback.

I've added materials parsing to the YAML format and material generation for the `to_urdf` and `to_yaml` functions.

I'll merge this and release a new version of RBDyn

PR #90, opened 2021-10-05
[export] Ensure that filenames are prefixed by file:// or package://

arntanguy:
This PR ensures that filenames in the `urdf`/`yaml` obtained by `to_urdf`/`to_yaml` are always prefixed by either `file://`  or `package://`. This is a requirement for a lot of ROS tools (such as rviz).

For example the following cannot be displayed in rviz
```
<mesh filename="/usr/local/share/panda_prosthesis/meshes/support_femur.stl" scale="0.001 0.001 0.001"/>
```
but with the `file://` prefix everything is fine.
```
<mesh filename="file:///usr/local/share/panda_prosthesis/meshes/support_femur.stl" scale="0.001 0.001 0.001"/>
```

Note that this assumes that in case no valid prefix is provided that the intent was to load a local file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-16.04, ubuntu-18.04, macos-latest, windows-latest]
+        os: [ubuntu-18.04, ubuntu-20.04, macos-latest, windows-latest]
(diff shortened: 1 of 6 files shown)


gergondet (changes requested):
Looks ok.

You need to update the reference URDF used in the tests: https://github.com/jrl-umi3218/RBDyn/blob/master/tests/ParsersTestUtils.h#L220

PR #96, opened 2023-04-26
[RBDyn/parsers] Parse mesh->scale as a vector

gergondet:
This PR also updates the cmake submodule to use the more resilient Cython bindings generation.

It also sets a pre-commit configuration and removes the clang-format related scripts in favor of the automated pre-commit.ci usage


mmurooka:
Hi @gergondet 

> It also sets a pre-commit configuration and removes the clang-format related scripts in favor of the automated pre-commit.ci usage

Do developers need any steps to install and set up pre-commit, especially in an Ubuntu 20.04 environment? What is the standard installation procedure for pre-commit?

gergondet:
Hi @mmurooka 

There's nothing to be done if you are using mc-rtc-superbuild as it's done automatically for you.

Other wise:
1. Install pre-commit (`python3 -m pip install pre-commit`)
2. Install the hooks in the repository that is using pre-commit: `pre-commit install`

mmurooka:
Thanks for the explanation @gergondet

PR #101, opened 2023-07-25
Do not export symbols when building static libs

costashatz:
Do not export symbols when building static library.. I guess this should be the correct way of doing this, no?

### DIFF
--- a/src/RBDyn/CMakeLists.txt
+++ b/src/RBDyn/CMakeLists.txt
@@ -53,7 +53,11 @@ target_include_directories(
          $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include>
          $<INSTALL_INTERFACE:include>)
 target_link_libraries(RBDyn PUBLIC SpaceVecAlg::SpaceVecAlg)
+if(BUILD_SHARED_LIBS)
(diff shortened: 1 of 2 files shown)


gergondet:
@costashatz 

What's the problem with the static build? Is it a problem to use `__declspec(dllexport)` when building a static library on Windows?

PS: I think the fix needs to be different since the `RBDYN_STATIC` flag needs to be adopted downstream to consume the API correctly in that case, it also needs to be applied to the parsers library

costashatz:
> What's the problem with the static build? Is it a problem to use `__declspec(dllexport)` when building a static library on Windows?

I have little experience with Windows, but in my project using `__declspec(dllexport)` to build the static lib and then `__declspec(dllimport)` to use it generates many `C4251` warnings (which I treat as errors).

> the `RBDYN_STATIC` flag needs to be adopted downstream to consume the API correctly in that case

What do you mean by that?

> it also needs to be applied to the parsers library

True! I am not using the parsers and forgot to add it. I will.

gergondet:
Thanks for the clarification @costashatz 

I have made some small changes to the patch (fixed the CMake formatting and the typo for RBDynParsers) and pushed that to the master branch.

PR #109, opened 2023-11-03
Relative jacobian fix

BenjaminNavarro:
Supersedes #78 

See both commit messages for the two individual fixes description

### DIFF
--- a/src/RBDyn/Jacobian.cpp
+++ b/src/RBDyn/Jacobian.cpp
@@ -20,8 +20,25 @@ namespace rbd
 Jacobian::Jacobian() {}
 
 Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eigen::Vector3d & point)
-: Jacobian(mb, bodyName, mb.body(0).name(), point)
+: jointsPath_(), point_(point), jac_(), jacDot_()
 {
+  bodyIndex_ = mb.sBodyIndexByName(bodyName);
+  refBodyIndex_ = 0;
+
+  int index = bodyIndex_;
+
+  int dof = 0;
+  while(index != -1)
+  {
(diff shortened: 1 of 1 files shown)


src/RBDyn/Jacobian.cpp
```diff
@@ -20,8 +20,25 @@ namespace rbd
 Jacobian::Jacobian() {}
 
 Jacobian::Jacobian(const MultiBody & mb, const std::string & bodyName, const Eigen::Vector3d & point)
-: Jacobian(mb, bodyName, mb.body(0).name(), point)
```
gergondet, line 23: I think this should still work, the first `while` loop in the constructor will break on `index = 0` so the `index != refBodyIndex_` branch is not taken and we end up with the same code

gergondet (changes requested):
Other than the point about the potentially duplicated constructor this looks ok 

Thanks for rebasing this on the latest master.

src/RBDyn/Jacobian.cpp
BenjaminNavarro: I ended up with the two constructors because I couldn't find an easy way to make the base case (no ref body) works after the second fix as it would always discard the model's root joint. I could use something like an empty string or a special name for the ref body to know that I need to go all the way to the root joint but I thought that in the end it was simpler and less error prone like that.

gergondet: Ah I see, I walked through the code while looking at the current version but with the modified version it will indeed skip the root body.

> I could use something like an empty string or a special name for the ref body to know that I need to go all the way to the root joint but I thought that in the end it was simpler and less error prone like that.

I think we can go with `if(refBodyIndex_ != 0 && index == refBodyIndex_)` as a condition to break?

PR #119, opened 2024-02-02
Remove Geometry::scale 

arntanguy:
This PR proposes to remove the duplication between `double Geometry::Mesh::scale` and `Eigen::Vector3d Geometry::Mesh::scaleV`. Indeed `scaleV` contains all the necessary information and having both is unnecessary and confusing.

Loading from `urdf/yaml` still supports both a single valued or dimensional scale. Exporting always exports the dimensional scale.

See https://github.com/jrl-umi3218/mc_rtc/pull/430


gergondet:
Hi @arntanguy 

Thanks for the PR I have no problem with it but I suggest we make sure that no code (that we are aware of) relies on `scale` first. Release new versions of the packages that did and finally merge and release a version of RBDyn that removes `scaleV`

arntanguy:
Yes, magnum does, but besides that I am not aware of any other that relies on it.

